PR #14, opened 2021-03-11
Expect gzip encoding in probeinfo service response

relud:


generator/namespaces.py
```diff
@@ -56,7 +57,7 @@ def namespaces(custom_namespaces, generated_sql_uri, app_listings_uri):
     get_app_name = itemgetter("app_name")
     # groupby requires input be sorted by key to produce one result per key
     with urllib.request.urlopen(app_listings_uri) as f:
-        app_listings = sorted(json.load(f), key=get_app_name)
+        app_listings = sorted(json.loads(gzip.decompress(f.read())), key=get_app_name)
```
fbertsch, line 60: Worth mentioning, while we're here, that we bypass the cloudfront cache in MSG: https://github.com/mozilla/mozilla-schema-generator/blob/main/mozilla_schema_generator/generic_ping.py#L214-L218

fbertsch (approved):

generator/namespaces.py
relud: i'll add that in

relud: done

PR #19, opened 2021-03-16
Differentiate stable and derived views

### LINKED ISSUE #13: Differentiate stable and derived views when generating namespaces.yaml
now available via the `references` section of `metadata.yaml`

relud:
fixes #13

### DIFF
--- a/generator/namespaces.py
+++ b/generator/namespaces.py
@@ -29,10 +29,16 @@ def _get_views(uri):
     views = defaultdict(dict)
     with tarfile.open(fileobj=tarbytes, mode="r:gz") as tar:
         for tarinfo in tar:
-            if tarinfo.name.endswith("/view.sql"):
+            if tarinfo.name.endswith("/metadata.yaml"):
+                metadata = yaml.safe_load(tar.extractfile(tarinfo.name))
+                references = metadata.get("references", {})
+                if "view.sql" not in references:
+                    continue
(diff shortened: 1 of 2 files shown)


generator/namespaces.py
```diff
@@ -74,15 +80,21 @@ def namespaces(custom_namespaces, generated_sql_uri, app_listings_uri):
...
-            for view_id in view_definitions[dataset_id]:
+            for view_id, references in view_definitions[dataset_id].items():
                 if view_id in OMIT_VIEWS:
                     continue
                 table = {"table": f"mozdata.{dataset_id}.{view_id}"}
                 if "app_channel" in app:
                     table["channel"] = app["app_channel"]
+                if len(references) == 1 and references[0][-2] == f"{dataset_id}_stable":
+                    # ping views reference a single table in the stable dataset
+                    table["is_ping"] = True
```
fbertsch, line 95: `is_ping` is a bit confusing to read without context in `namespaces.yaml`. What do you think of doing the inverse, and calling it `is_derived_table`?

fbertsch (approved):
r+wc

generator/namespaces.py
relud: I don't like `is_derived_table` because I don't think we should actually categorize everything else as "derived" (e.g. we also have `_static`, `_external`, etc)

per slack, let's go with `is_ping_table`

PR #22, opened 2021-03-22
Namespaces explores

fbertsch:
This does nothing yet except expose a new section of `namespaces.yaml` called `explores`. Eventually we'll do a few things:

1. Just use this section to generate LookML, and not need to decipher which explores we need from the views
2. Generate LookML within the explore classes (extensibly, using the dimensions)

### DIFF
--- a/generator/explores.py
+++ b/generator/explores.py
@@ -1,0 +1,34 @@
+"""All possible generated explores."""
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Dict, Iterator, List
+
+
+@dataclass
+class Explore:
+    """A generic explore."""
+
+    name: str
+    type: str
+    views: Dict[str, str]
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ Ensure Python 3.8+ is available on your machine (see [this guide](https://docs.p
 
 Install requirements in a Python venv
 ```bash
-python3.8 -m venv venv/
+python3.8 -m virtualenv venv/
```
relud, line 32: this intentionally uses `venv` l, which is built in, instead of `virtualenv`, which has to be installed

relud (approved):
r+ for everything except `virtualenv`

README.md
fbertsch, line 32:
```suggestion
python3.8 -m venv venv/
```

PR #24, opened 2021-03-22
Generate model for each namespace

### LINKED ISSUE #20: Generate model lookml
This should probably look something like `{app_name}/{app_name}.model.lkml`:
```lkml
connection: "telemetry"
label: {canonical_app_name}

include "views/*.view"
include: "dashboards/*.dashboard"

explore: {view_name} {
  sql_always_where: ${submission_raw} >= "2020-01-01" ;;
}

...
```

with explores for ping views only ~but when I try to validate this in looker I get errors like `Model 'fenix' has not been configured`~

relud:
fixes #20

### DIFF
--- a/generator/__init__.py
+++ b/generator/__init__.py
@@ -1,5 +1,6 @@
 """Generate LookML."""
 
+import logging
 import warnings
 
 import click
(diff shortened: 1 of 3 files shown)


fbertsch (changes requested):
This is generally correct but we don't actually want to generate LookML models in the hub project. The first time LookML is generated for a namespace, we will create an entry for it in the spoke project. Subsequent runs will not update that file, since it could have been edited by developers at that point. That model will import the hub explores (I do not believe it needs to import the hub views, but we obvs. need to verify that).

PR #26, opened 2021-03-22
Generate views via lkml

relud:


generator/lookml.py
```diff
@@ -41,196 +37,140 @@
...
-        try:
-            measures = Measure.from_dimensions(dimensions)
-        except KeyError as e:
-            raise click.ClickException(f"duplicate measure {e} for table {table!r}")
-        fields = []
-        sql_table_name = table
+        # add dimensions and dimension groups
+        dimensions = _generate_dimensions(client, table)
+        view["dimensions"] = [d for d in dimensions if "timeframes" not in d]
+        view["dimension_groups"] = [d for d in dimensions if "timeframes" in d]
```
fbertsch, line 151: I think we need a more robust method of determining if a dimension is a dimension_group ([docs](https://docs.looker.com/reference/field-params/dimension_group#intervals)). Reading that, it should be that either `timeframes` is in d or `intervals` is in d. Maybe we should pull this to a separate function, e.g.
```
view["dimensions"] = filter(d, lambda x: not is_dimension_group(x))
view["dimension_groups"] = filter(d, is_dimension_group)
```

fbertsch (approved):
r+wc. This should be more maintainable and extensible

generator/lookml.py
relud: done, with one difference: used `filterfalse` from the standard library's `itertools`

PR #27, opened 2021-03-24
View classes

fbertsch:
The output right now is exactly the same as before. Next step will be to expand the GrowthAccountingView to create a view, and then an explore.

### DIFF
--- a/generator/namespaces.py
+++ b/generator/namespaces.py
@@ -47,8 +47,8 @@ def _get_db_views(uri):
 def _get_looker_views(
     app_name: str, variants: dict, db_views: Dict[str, Dict[str, List[List[str]]]]
 ):
-    views = defaultdict(list)
-    for _, klass in view_types.items():
+    views = {}
+    for klass in view_types.values():
         for view_id, tables in klass.from_db_views(variants, db_views).items():
             if view_id in views:
                 raise KeyError(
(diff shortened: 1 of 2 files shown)


generator/explores.py
```diff
@@ -43,4 +43,26 @@ def from_dict(name: str, defn: dict) -> PingExplore:
         return PingExplore(name, "ping_explore", defn["views"])
 
 
-explore_types = {"ping_explore": PingExplore}
+@dataclass
```
fbertsch, line 46: Will remove this from this PR.

generator/namespaces.py
```diff
@@ -84,35 +101,21 @@ def namespaces(custom_namespaces, generated_sql_uri, app_listings_uri):
     with urllib.request.urlopen(app_listings_uri) as f:
         # groupby requires input be sorted by key to produce one result per key
         app_listings = sorted(json.loads(gzip.decompress(f.read())), key=get_app_name)
-    view_definitions = _get_views(generated_sql_uri)
+    db_views = _get_db_views(generated_sql_uri)
     namespaces = {}
     for app_name, group in groupby(app_listings, get_app_name):
-        views = defaultdict(list)
+        group = list(group)
         canonical_app_name = None
```
relud, line 108: I think this should reuse the pattern from https://github.com/mozilla/lookml-generator/blob/5a702a7d385ab155e8c4a52c6ea8df84a292d0fd/generator/lookml.py#L149-L153

```suggestion
        # use canonical_app_name where channel=="release" or the first one
        canonical_app_name = next(
            (app for app in group if app.get("channel") == "release"),
            group[0],
        )["canonical_app_name"]
```

generator/namespaces.py
```diff
@@ -45,6 +44,24 @@ def _get_views(uri):
     return views
 
 
+def _get_looker_views(
+    app_name: str, variants: dict, db_views: Dict[str, Dict[str, List[List[str]]]]
+):
+    views = defaultdict(list)
```
relud, line 50: Using a `defaultdict` here is misleading because the default is never used, an error is thrown before a value could be assigned more than once, and the result is immediately cast to a regular `dict`
```suggestion
    views = {}
```

relud (approved):
r+wc

generator/namespaces.py
fbertsch: Much better

generator/namespaces.py
fbertsch: I agree, it was a holdover from the previous version of this code.

PR #30, opened 2021-04-16
Ga explore lookml

fbertsch:


generator/lookml.py
```diff
@@ -1,211 +1,50 @@
 """Generate lookml from namespaces."""
 import logging
-import re
-from itertools import filterfalse
 from pathlib import Path
-from typing import Any, Dict, Iterable, List, Tuple, TypedDict
+from typing import Dict, Iterable
```
fbertsch, line 4: Moved most of the stuff in this class to either the View definition (for things that are specific to the view type, e.g. Measures) or `lookml_utils.py`, for the dimension generation (since that is the same for any view, in theory).

generator/views.py
```diff
@@ -91,11 +119,289 @@ def from_db_views(
...
+            ),
+            "type": "yesno",
+            "hidden": "yes",
+        },
+    ]
+
+    default_measures: List[Dict[str, Union[str, Dict[str, str]]]] = [
+        {
+            "name": "overall_active_previous",
+            "type": "count",
```
relud, line 253: either this view needs to define `client_id` as a primary key, or all these count measures need to use `count_distinct`, e.g.:
```suggestion
            "type": "count_distinct",
            "sql": "client_id",
```

tests/test_lookml.py
```diff
@@ -406,6 +478,8 @@ def test_duplicate_measure(runner, tmp_path):
     )
     with runner.isolated_filesystem():
         with patch("google.cloud.bigquery.Client", MockClient):
+            # print(f"{namespaces}")
+            # lookml(str(namespaces), "looker-hub/")
```
relud, line 482:
```suggestion
```

relud (approved):
r+wc

generator/views.py
```diff
@@ -91,11 +119,289 @@ def from_db_views(
...
+        for dimension in dimensions:
+            dimension_name = dimension["name"]
+            if dimension_name in {"client_id", "client_info__client_id"}:
+                measure = {
+                    "name": "clients",
+                    "type": "count_distinct",
+                    "sql": f"${{{dimension_name}}}",
+                }
+            elif dimension_name == "document_id":
+                measure = {"name": "ping_count", "type": "count"}
```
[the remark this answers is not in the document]
fbertsch: It should be fine to keep this as a `count`, since we don't have any joins yet. Once we enable joins we can certainly add `document_id` as the primary key.

generator/views.py
fbertsch: We definitely want to avoid `count_distinct` for these measures, where possible. I'm fine with adding `client_id` - `submission_date` as primary key.

Issue #36, opened 2021-04-19
Create CLI command to generate folder


fbertsch:
This follows up from the [content and dashboard organization proposal](https://docs.google.com/document/d/1UeM4icJkBc0t-bbrY8obG-4sKn7I-5RYQZhKJfWXdqI/edit#).

For every namespace, we need to ensure:
- There are users associated with that namespace
- There is a folder in the shared folder for that namespace
- The users associated with that namespace have write access to that folder
- Admins have write access to that folder
- Other users have read access to that folder

PR #43, opened 2021-04-20
Content folders

fbertsch:


fbertsch:
Fixes https://github.com/mozilla/lookml-generator/issues/36.

Would have loved to use terraform instead.

generator/content.py
```diff
@@ -0,0 +1,146 @@
...
+
+    shared_folders = sdk.search_folders(name="home")
+    if len(shared_folders) != 1:
+        raise click.ClickException(
+            f"Found wrong number of shared folders: {shared_folders}"
+        )
+
+    shared_folder_id = shared_folders[0].id
+    if shared_folder_id is None:
+        raise click.ClickException("Error: Shared folder id is missing")
```
relud, line 36: this pattern is used so many times, I think it should be a separate function that is passed a list and returns an id

relud (approved):
r+wc

generator/content.py
fbertsch: Makes sense, thanks for the suggestion!

PR #45, opened 2021-04-20
Update spoke repo in bin/generate

fbertsch:
- Pull out common operations to function (for comitting)
- Makes the hub publish branch exactly match the source, rather
  than cherry-picking

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,6 @@ a new branch to the `looker-hub` repository:
 
 ```
 export HUB_BRANCH_PUBLISH="yourname-generation-test-1"
-export HUB_SSH_KEY_BASE64=$(cat ~/.ssh/id_rsa | base64)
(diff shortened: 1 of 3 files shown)


fbertsch:
@acmiyaguchi recent test here: https://github.com/mozilla/looker-spoke-default/tree/test-lookml-generation
You can see that the recent commit dropped a bunch of fields, which we want, since only `burnham` is in the namespace allowlist.

fbertsch:
Er, I misspoke earlier. That is pointing to the spoke project, which should only ever have new directories. [This commit](https://github.com/mozilla/looker-hub/tree/test-lookml-generation) to looker-hub was the one that deleted a bunch of files (expectedly).

acmiyaguchi (approved):

PR #49, opened 2021-04-21
Configure model after creation

fbertsch:
We've discovered on stage that the models are not configured after deploy.

We add that here to ease development lifecycle. I've tested on dev and this looks to be successful.

Webhooks ensure that the instance is up-to-date on production.


fbertsch:
Not yet ready for review. It seems this isn't working as expected on dev.

relud (approved):
r+wc

relud (approved):

PR #52, opened 2021-04-26
Open PR for spoke changes

fbertsch:
- Remove GCLOUD_SERVICE_KEY as unnecessary

This matches what we discussed on Friday, @jasonthomas. Let me know if this method of secret passing (I used a GH personal access token) won't work.


bin/generate
```diff
@@ -215,13 +221,24 @@ function hit_looker_webhooks() {
   curl "$LOOKER_INSTANCE_URI/webhooks/projects/spoke-default/deploy"
 }
 
+function open_spoke_pull_request() {
+  # Open a PR to merge the changes from working branch
+  # in to the publish branch.
+  gh pr create \
+    --title "Auto-push from LookML Generator" \
+    --reviewer fbertsch \
```
fbertsch, line 229: It would be good to have a GH team that this can hit as reviewer instead.

jasonthomas: I can create a new team, please provide me with the members. Alternately we can use CODEOWNERS and specify individuals in it. I believe it will automatically force CODEOWNERS as reviewers on PR requests [1].

[1] https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

jasonthomas (approved):
Personal access token should be fine.

bin/generate
fbertsch: I don't think we should use code owners, since that would require review from one of the members on non-generated LookML PRs (as far as I can tell). Let's create a new team and start with the members of the working group: You, me, relud, and anthony.

jasonthomas: Created team https://github.com/orgs/mozilla/teams/data-looker

fbertsch: Great, that [worked as expected](https://github.com/mozilla/looker-spoke-default/pull/30).

PR #58, opened 2021-04-27
Generate glean dictionary links for metric columns in pings (fixes #64)

### LINKED ISSUE #64: Add links to Glean dictionary for Glean Ping explores
This should roughly match the [unified metrics](https://github.com/mozilla/looker-spoke-default/blob/main/firefox_ios/views/unified_metrics.view.lkml) view. All dimensions and measures that have an entry in the Glean dictionary should have a link to that entry.

wlach:
This is horrible and hacky, but would appreciate some feedback if I'm on the right track. It does seem to yield valid results in my initial (very cursory) testing.

### DIFF
--- a/generator/explores/__init__.py
+++ b/generator/explores/__init__.py
@@ -1,9 +1,17 @@
 """All possible explore types."""
 from .explore import Explore  # noqa: F401
+from .glean_ping_explore import GleanPingExplore
 from .growth_accounting_explore import GrowthAccountingExplore
 from .ping_explore import PingExplore
 
 EXPLORE_TYPES = {
+    GleanPingExplore.type: GleanPingExplore,
     PingExplore.type: PingExplore,
     GrowthAccountingExplore.type: GrowthAccountingExplore,
+}
+
+GLEAN_EXPLORE_TYPES = {
+    t: v
+    for (t, v) in EXPLORE_TYPES.items()
+    if t in [GleanPingExplore.type, GrowthAccountingExplore.type]
 }
--- a/generator/explores/__init__.py
+++ b/generator/explores/__init__.py
@@ -9,9 +9,3 @@ EXPLORE_TYPES = {
     PingExplore.type: PingExplore,
     GrowthAccountingExplore.type: GrowthAccountingExplore,
 }
-
-GLEAN_EXPLORE_TYPES = {
-    t: v
-    for (t, v) in EXPLORE_TYPES.items()
-    if t in [GleanPingExplore.type, GrowthAccountingExplore.type]
-}
--- a/generator/explores/glean_ping_explore.py
+++ b/generator/explores/glean_ping_explore.py
@@ -1,7 +1,8 @@
+"""Glean Ping explore type."""
 from typing import Iterator, List
 
-from .ping_explore import PingExplore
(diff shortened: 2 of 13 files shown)


wlach:
Example lookml: https://gist.github.com/wlach/e2cc042fa3f32f1b0ea0600417aae7e7

generator/views/lookml_utils.py
```diff
@@ -65,20 +67,33 @@ def _get_dimension(path: Tuple[str, ...], field_type: str, mode: str) -> Dict[st
         if path in MAP_LAYER_NAMES:
             result["map_layer_name"] = MAP_LAYER_NAMES[path]
     result["name"] = "__".join(name)
+    if name[0] == "metrics":
```
wlach, line 70: We could probably cover client_info metrics quite easily with this approach too. It is admittedly a huge hack job.

fbertsch:
Okay, looking this over, I think we need to go in a different direction. My idea is basically this: Let's create a separate view and explore type called `GleanPing`. Then we can pass the namespace in and pull down the probe-info-service data. That would allow us to:
- Add these Glean Dictionary integration points
- Include descriptions from probeinfo.tmo
- Create [per-category dropdowns as dimensions](https://github.com/mozilla/lookml-generator/issues/61)

wlach:
@fbertsch This sounds great, is this something you'd like me to implement? I have some bandwidth to work on this but there are also a few other things I need to take care of today / tomorrow.

generator/explores/__init__.py
```diff
@@ -1,9 +1,17 @@
...
+    GleanPingExplore.type: GleanPingExplore,
     PingExplore.type: PingExplore,
     GrowthAccountingExplore.type: GrowthAccountingExplore,
 }
+
+GLEAN_EXPLORE_TYPES = {
+    t: v
+    for (t, v) in EXPLORE_TYPES.items()
+    if t in [GleanPingExplore.type, GrowthAccountingExplore.type]
+}
```
fbertsch, line 17: I don't think we have to go the path of explicitly separating these anymore - I added `glean_app: $bool` to the `namespaces.yaml` entries, we can check in the namespace and only return an instance of the view/explore if it is a Glean application.

wlach: Yeah, this was a relic of my work on Friday that I forgot to clean up. I'll take it out.

wlach: Actually I think `GLEAN_VIEW_TYPES` might be the most elegant way of making sure we don't generate duplicate views for the same Glean ping. I did take out `GLEAN_EXPLORE_TYPES` though.

wlach: Specifically: https://github.com/mozilla/lookml-generator/pull/58/files#diff-a6989594893a2eff2aa73d1affe2a11cc1a85effed9ef992a025847501788140R107

fbertsch: Mmm, Let's discuss more later, but in general I prefer the logic of  "Does this definition generate a view for this table" to live in one place.

fbertsch (approved):
Fixed this up, and merging in. Tested here: https://github.com/mozilla/looker-hub/compare/frank-test-generation-3?expand=1

PR #62, opened 2021-04-28
Add Fenix to allowlist

fbertsch:
I'm hesitant about merging this change just yet. I'm noticing some things about the explores that may not be backwards compatible if we enact them, and we are blocked from further stage deploys if we merge this as-is.

### DIFF
--- a/namespace_allowlist.yaml
+++ b/namespace_allowlist.yaml
@@ -1,2 +1,3 @@
 ---
 - burnham
+- fenix


fbertsch:
Available on dev, e.g.: https://mozilladev.cloud.looker.com/explore/fenix/metrics

fbertsch:
Added in separate PR.

PR #66, opened 2021-04-28
Updates to `namespaces.yaml`

fbertsch:
- Use pretty_name instead of canonical_app_name
- Add `glean_app` to indicate a namespace is glean

### DIFF
--- a/generator/content.py
+++ b/generator/content.py
@@ -49,7 +49,7 @@ def generate_folders(namespaces: dict):
     shared_folder_id = _get_id_from_list(shared_folders, "Shared folders")
 
     for namespace, defn in namespaces.items():
-        pretty_name = defn["canonical_app_name"]
+        pretty_name = defn["pretty_name"]
         owners = defn["owners"]
 
         try:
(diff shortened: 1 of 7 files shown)


wlach (changes requested):
Less of a "request changes" than "request changes or explain rationale"

Is there a particular reason you prefer `pretty_name` to `canonical_app_name`? I can see how it's more intuitive what the field means for someone looking at Looker by itself, but `canonical_app_name` is nice because we can draw a connection back to what we define in probe-scraper: https://mozilla.github.io/probe-scraper/#tag/library

fbertsch:
> Less of a "request changes" than "request changes or explain rationale"
> 
> Is there a particular reason you prefer `pretty_name` to `canonical_app_name`? I can see how it's more intuitive what the field means for someone looking at Looker by itself, but `canonical_app_name` is nice because we can draw a connection back to what we define in probe-scraper: https://mozilla.github.io/probe-scraper/#tag/library

Good point - basically because `canonical_app_name` doesn't mean anything for custom namespaces (consider e.g. numbers that matter, revenue, etc.). IMO we should be using the namespace name to map to the `app_name`, when `glean_app` is true.

wlach (approved):
Makes sense! Please go ahead and merge

PR #69, opened 2021-04-28
Add numbers_that_matter namespace

scholtzan:
I'm not entirely sure what the process looks like here. Once this gets merged, will the namespace be automatically generated and views will be added to looker-hub? Or do I have to open a PR against looker-hub to add what was generated?

The generated output: https://github.com/scholtzan/looker-hub/tree/ascholtz-numbers-that-matter-generation-test-1

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1,1 +1,17 @@
 ---
+duet:
+  glean_app: false
+  owners:
+  - ascholtz@mozilla.com
+  pretty_name: DUET
+  views:
+    install:
+      type: ping_view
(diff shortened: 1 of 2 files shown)


fbertsch (approved):
This looks good! You don't have to anything else here, except we will need to tag a new release of `lookml-generator`. Basically that will run and create everything for you.

Question: Should "Numbers that Matter" instead be an explore/dashboard inside a separate namespace? e.g. should the namespace be for `DUET`?

scholtzan:
Good point! I think having a DUET namespace does make sense

PR #71, opened 2021-04-28
Require date and channel filters in ping explores

### LINKED ISSUE #59: Require channel selector for explores that have them
If the channel selector isn't selected, LookML doesn't include any table at all. To fix this, we can have the explore set a default value for that field.

### LINKED ISSUE #60: Require date selection for Ping Explores
Currently, if you query a ping explore, it will fail unless you add a date selection. This is because the underlying tables require a date filter.

We can add a required date filter that defaults to a sane value - e.g. the past month.

fbertsch:
Only requires a channel when it is present in the view. Fixes #59, Fixes #60.

### DIFF
--- a/tests/test_lookml.py
+++ b/tests/test_lookml.py
@@ -159,7 +159,7 @@ def test_lookml_actual(runner, tmp_path):
     )
     with runner.isolated_filesystem():
         with patch("google.cloud.bigquery.Client", MockClient):
-            _lookml(Path(namespaces.absolute()).read_text(), "looker-hub/")
+            _lookml(open(namespaces), "looker-hub/")
         expected = {
             "views": [
                 {
--- a/tests/test_lookml.py
+++ b/tests/test_lookml.py
@@ -1,4 +1,3 @@
-import sys
 from pathlib import Path
 from textwrap import dedent
 from unittest.mock import patch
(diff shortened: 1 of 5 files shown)


tests/test_lookml.py
```diff
@@ -160,21 +159,7 @@ def test_lookml_actual(runner, tmp_path):
...
-            )
-        sys.stdout.write(result.stdout)
-        if result.stderr_bytes is not None:
-            sys.stderr.write(result.stderr)
-        try:
-            assert result.exit_code == 0
-        except Exception as e:
-            # use exception chaining to expose original traceback
-            raise e from result.exception
+            _lookml(Path(namespaces.absolute()).read_text(), "looker-hub/")
```
fbertsch, line 162: Switched the way we run tests here. I was running in to problems where print statements cause test failures and logs are eaten up.

relud, line 162: `namespaces` as provided by `click` should be an open file pointer:
```suggestion
            _lookml(open(namespaces), "looker-hub/")
```

relud (approved):
r+wc

tests/test_lookml.py
fbertsch: I think I tried that... let me commit this suggestion and see

PR #75, opened 2021-04-30
Add generated models to user_spokes model set

fbertsch:


jasonthomas (approved):

generator/spoke.py
```diff
@@ -13,10 +13,10 @@
...
 from .lookml import ViewDict
 
-MODEL_SETS_BY_INSTANCE = {
-    "https://mozilladev.cloud.looker.com": "mozilla_confidential",
-    "https://mozillastaging.cloud.looker.com": "mozilla_confidential",
-    "https://mozilla.cloud.looker.com": "user-spokes",
+MODEL_SETS_BY_INSTANCE: Dict[str, List[str]] = {
+    "https://mozilladev.cloud.looker.com": ["mozilla_confidential"],
+    "https://mozillastaging.cloud.looker.com": ["mozilla_confidential"],
+    "https://mozilla.cloud.looker.com": ["user-spokes", "spokes"],
```
jasonthomas, line 19: We eventually should migrate to what is proposed in the data access management proposal for prod, but this is okay for now.

fbertsch: Yeah, that's why I tagged you for review. We'll need to first deprecate the `spokes` project, then we can move all the models to `mozilla_confidential` instead of `user-spokes`, then finally we can deprecate `user-spokes`. (`spokes` currently has a few models that are not available to `mozilla_confidential`).

PR #78, opened 2021-04-30
Add sums and client counts for `metrics.counter`

acmiyaguchi:
This adds some measures for counters in the glean pings. It looks like there will probably more general ways of handling glean pings, but this is a start.

Here's an example push: https://github.com/mozilla/looker-hub/commit/68cd4f805e068d64f1a841b7a280eb58864fb54a

### DIFF
--- a/generator/views/ping_view.py
+++ b/generator/views/ping_view.py
@@ -1,7 +1,7 @@
 """Class to describe a Ping View."""
 from __future__ import annotations
 
-from collections import defaultdict
+from collections import Counter, defaultdict
 from itertools import filterfalse
 from typing import Any, Dict, Iterator, List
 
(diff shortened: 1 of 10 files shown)


generator/views/ping_view.py
```diff
@@ -93,37 +93,68 @@ def to_lookml(self, bq_client) -> List[dict]:
...
+
+        # check if there are any duplicate values, and report the first one that
+        # shows up
+        names = [measure["name"] for measure in measures]
+        duplicates = [k for k, v in Counter(names).items() if v > 1]
+        if duplicates:
+            name = duplicates[0]
             raise click.ClickException(
-                f"Missing client_id and doc_id dimensions in {table!r}"
+                f"duplicate measure {name!r} for table {table!r}"
```
acmiyaguchi, line 157: I can't actually think of any situations where this would occur, outside of poorly specifying the loop above (e.g. only using `dimension_name[-1]`).

fbertsch (commented):
This is a good start! I think we should see what @wlach has up his sleeve and build on the new `GleanPing` view, but it should roughly match what you've done here. We will also enable it for all Glean pings, not just metrics pings. The three of us should catch up on Monday to discuss.

acmiyaguchi:
I'll be out all of next week, but I can rebase (or redo) any of the work here to support the counters. I also want to try my hands at the labeled counters too.

fbertsch (approved):
Tested in https://github.com/mozilla/looker-hub/commit/b3a1e7adb7a0883552dc8be0a9f6712681b8cd2a.

PR #79, opened 2021-05-03
Expose Firefox iOS and add `TableView`

fbertsch:
A TableView simply exposes a raw table, with no measures.

Eventually we will import this instead of re-generating
all dimensions multiple times.

@wlach I think this may conflict with your changes. Take a look and let me know, I'd rather not block you, and can rebase on yours once we merge that.

### DIFF
--- a/GENERATION_DISALLOW_LIST
+++ b/GENERATION_DISALLOW_LIST
@@ -1,3 +1,8 @@
 firefox_ios/views/baseline.view.lkml
 firefox_ios/views/events.view.lkml
 firefox_ios/views/metrics.view.lkml
+firefox_ios/views/growth_accounting.view.lkml
+firefox_ios/explores/events.explore.lkml
+firefox_ios/explores/growth_accounting.explore.lkml
+firefox_ios/explores/metrics.explore.lkml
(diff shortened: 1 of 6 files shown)


fbertsch:
Tested the changes to `bin/generate` via https://github.com/fbertsch/looker-hub/commit/2a395606efa5bb47105435cbdd6ed3ced8277bdc

GENERATION_DISALLOW_LIST
```diff
@@ -1,3 +1,8 @@
 firefox_ios/views/baseline.view.lkml
 firefox_ios/views/events.view.lkml
 firefox_ios/views/metrics.view.lkml
+firefox_ios/views/growth_accounting.view.lkml
+firefox_ios/explores/events.explore.lkml
+firefox_ios/explores/growth_accounting.explore.lkml
+firefox_ios/explores/metrics.explore.lkml
+firefox_ios/explores/baseline.explore.lkml
```
fbertsch, line 8: Basically, we're not going to expose much generated LookML right now for Firefox iOS. We can always add more later.

relud (approved):

PR #86, opened 2021-05-07
Add Client Counts view and explore

fbertsch:
- Add updated lkml parser to address https://github.com/joshtemple/lkml/issues/53

### DIFF
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -7,6 +7,8 @@ from typing import Dict, List, Optional
 
 import lkml
 
+from ..views.lookml_utils import escape_filter_expr
+
 
 @dataclass
 class Explore:
@@ -27,7 +29,11 @@ class Explore:
 
     def get_dependent_views(self) -> List[str]:
         """Get views this explore is dependent on."""
-        return [view for _, view in self.views.items()]
+        return [
+            view
+            for _type, view in self.views.items()
+            if not _type.startswith("extended")
+        ]
 
     @staticmethod
     def from_dict(name: str, defn: dict, views_path: Path) -> Explore:
@@ -39,3 +45,50 @@ class Explore:
         if self.views_path is not None:
             return lkml.load((self.views_path / f"{view}.view.lkml").read_text())
         raise Exception("Missing view path for get_view_lookml")
+
+    def _get_default_channel(self, view: str) -> Optional[str]:
+        channel_params = [
+            param
+            for _view_defn in self.get_view_lookml(view)["views"]
+            for param in _view_defn.get("parameters", [])
+            if _view_defn["name"] == view and param["name"] == "channel"
+        ]
+
+        if channel_params:
+            allowed_values = channel_params[0]["allowed_values"]
+            default_value = next(
+                (value for value in allowed_values if value["label"] == "Release"),
+                allowed_values[0],
+            )["value"]
+
+            return escape_filter_expr(default_value)
+        return None
+
+    def _get_view_has_submission(self, view: str) -> bool:
+        return (
+            len(
+                [
+                    dim
+                    for _view_defn in self.get_view_lookml(view)["views"]
+                    for dim in _view_defn.get("dimension_groups", [])
+                    if _view_defn["name"] == view and dim["name"] == "submission"
+                ]
+            )
+            > 0
+        )
+
+    def get_required_filters(self, view_name: str) -> List[Dict[str, str]]:
+        """Get required filters for this view."""
+        filters = []
+        view = self.views[view_name]
+
+        # Add a default filter on channel, if it's present in the view
+        default_channel = self._get_default_channel(view)
+        if default_channel is not None:
+            filters.append({"channel": default_channel})
+
+        # Add submission filter, if present in the view
+        if self._get_view_has_submission(view):
+            filters.append({"submission_date": "28 days"})
+
+        return filters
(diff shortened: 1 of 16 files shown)


fbertsch:
Tested this in dev, on the branches `test-client-counts` for both hub and spoke. Everything seemed to be in order.

generator/lkml_update.py
```diff
@@ -0,0 +1,49 @@
...
+            value: A string, tuple, or list to serialize
+        Raises:
+            TypeError: If input value is not of a valid type
+        Returns:
+            A generator of serialized string chunks
+        """
+        if isinstance(value, str):
+            return self.parse_pair(key, value)
+        elif isinstance(value, (list, tuple)):
+            if self.is_plural_key(key) and not self.parent_key == "query":
```
fbertsch, line 37: This line is the only one I changed, where I prevent it from expanding the list if the parent is a query. Issue is filed upstream.

generator/explores/explore.py
```diff
@@ -39,3 +45,50 @@ def get_view_lookml(self, view: str) -> dict:
         if self.views_path is not None:
             return lkml.load((self.views_path / f"{view}.view.lkml").read_text())
         raise Exception("Missing view path for get_view_lookml")
+
+    def _get_default_channel(self, view: str) -> Optional[str]:
+        channel_params = [
+            param
+            for _view_defn in self.get_view_lookml(view)["views"]
+            for param in _view_defn.get("parameters", [])
+            if _view_defn["name"] == view and param["name"] == "channel"
```
wlach, line 54: I wonder if this pattern should be abstracted out at some point into a helper method? May be premature to do it now though.

generator/lkml_update.py
wlach: Can you add a reference to the issue you filed in a comment, both for this method and in the body of the class? That way someone in the future can reason about why this exists without digging into history.

generator/lookml.py
```diff
@@ -41,8 +42,9 @@ def _generate_explores(
             ],
             "explores": [explore.to_lookml()],
         }
+        print(file_lookml)
```
wlach, line 45: Debugging?

```suggestion
```

generator/views/client_counts_view.py
```diff
@@ -0,0 +1,132 @@
...
+                  INTERVAL ${client_counts.weeks_since_first_seen} WEEK
+                {% elsif client_counts.months_since_first_seen._is_selected %}
+                  INTERVAL ${client_counts.months_since_first_seen} MONTH
+                {% elsif client_counts.years_since_first_seen._is_selected %}
+                  INTERVAL ${client_counts.months_since_first_seen} YEAR
+                {% endif %}
+              ) < current_date
+              """,
+        }
+    ]
```
wlach, line 58: I do wonder if python code is the best place to store these bits of lookml. We can figure this out at a later date of course.

generator/views/client_counts_view.py
```diff
@@ -0,0 +1,132 @@
...
+        namespace: str,
+        is_glean: bool,
+        channels: List[Dict[str, str]],
+        db_views: dict,
+    ) -> Iterator[ClientCountsView]:
+        """Get Client Count Views from db views and app variants."""
+        dataset = next(
+            (channel for channel in channels if channel.get("channel") == "release"),
+            channels[0],
+        )["dataset"]
```
wlach, line 94: Are we sure `channels` always has at least one element?

generator/lkml_update.py
```diff
@@ -0,0 +1,49 @@
+"""An updated lkml parser to handle explore queries."""
+from typing import List, Union
+
+from lkml.keys import KEYS_WITH_NAME_FIELDS
+from lkml.simple import DictParser
+from lkml.tree import BlockNode, DocumentNode, ListNode, PairNode
+
+
+def dump(obj: dict) -> str:
```
wlach, line 9: Can you give this a slightly more descriptive name, like `lkml_dump`? It's hard to see where this is coming from in the consumers.

wlach (approved):
No major concerns from a technical level, though I am not very familiar with the growth accounting methodologies-- you may want to get someone else to check over your work there at some point.

generator/views/client_counts_view.py
fbertsch: I think so, since we're storing them as native python objects. That big SQL string is a string in LookML as well, see https://github.com/mozilla/looker-spoke-default/blob/main/firefox_ios/views/client_counts.view.lkml#L23-L45 - that is, moving it to a separate representation would _still_ have to be a string.

generator/views/client_counts_view.py
fbertsch: It does, this comes from the [`variants`](https://github.com/mozilla/lookml-generator/blob/main/generator/namespaces.py#L62), so if there were no channels the app wouldn't exist. For example, Burnham has no "channels" per-se, but with a single variant it ends up as just `[{"channel": None, "dataset": "burnham"}]`.

This is probably not super clear from the code structure, so we should consider cleaning it up (but this problem exist in `probe-info-service` as well).

generator/lkml_update.py
fbertsch: What if instead we import the file? Then use `lkml_update.dump`.

generator/views/client_counts_view.py
wlach: May be worth a quick comment then.

PR #91, opened 2021-05-10
Bug 1709504 - Remove burnham from allowlist

fbertsch:


wlach (approved):
Do we have a long term plan for what will be and what won't be available in Looker? We have many Glean applications.

fbertsch:
Right now we're focused on the high-priority Glean applications. We'll tackle the longer tail after in H2. Details are in the mana page: https://mana.mozilla.org/wiki/display/DATA/Looker+Deploy+Working+Group

Issue #55, opened 2021-04-27
Add optional, additional owners to Glean apps


fbertsch:
For custom apps, we can have owners be a separate field.

For Glean apps, we should have a separate listing of owners here, that we can use to build out more owners of the namespace folder in Looker. These owners will have control of the dashboards in the shared folder.

PR #93, opened 2021-05-11
Add optional owners section to namespace_allowlist

fbertsch:
Closes https://github.com/mozilla/lookml-generator/issues/55.


generator/content.py
```diff
@@ -151,7 +151,13 @@ def generate_folders(namespaces: dict):
     type=click.File(),
     help="Path to a yaml namespaces file",
 )
-def generate_content(namespaces):
+@click.option(
+    "--allowlist",
+    default="namespace_allowlist.yaml",
```
relud, line 156: I think this should be named `namespaces-allowlist.yaml`, because `namespaces` is consistently plural elsewhere, and because `custom-namespaces.yaml` is already using kebab-case instead of snake_case for file names.
```suggestion
    default="namespaces-allowlist.yaml",
```

generator/content.py
```diff
@@ -40,7 +40,7 @@ def _get_id_from_list(arr: Sequence[Any], item: str) -> int:
     return int(first.id)
 
 
-def generate_folders(namespaces: dict):
+def generate_folders(namespaces: dict, addl_owners: dict = {}):
```
relud, line 43: I think that for consistency this should be named `allowlist`
```suggestion
def generate_folders(namespaces: dict, allowlist: dict = {}):
```

generator/content.py
```diff
@@ -50,7 +50,7 @@ def generate_folders(namespaces: dict):
 
     for namespace, defn in namespaces.items():
         pretty_name = defn["pretty_name"]
-        owners = defn["owners"]
+        owners = defn["owners"] + addl_owners.get(namespace, {}).get("owners", [])
```
relud, line 53:
```suggestion
        owners = defn["owners"] + allowlist.get(namespace, {}).get("owners", [])
```

relud (approved):
r+wc

generator/content.py
fbertsch: I agree - I'll update the filename.

relud (approved):

Issue #90, opened 2021-05-07
Make filter suggestions work for all tables


fbertsch:
To make filter suggestions work, we need to make two changes:
1. All parameterized table names (channels) need to have a `default_value`. This should be the Release channel value, if possible; but it should not be escaped with `^` (since it is not being used as a filter expression).
2. All explores need a `sql_always_where: "DATE(submission_timestamp) >= '2010-01-01';;"`, (or submission_date if that's what the table has). This will add a date filter to the filter suggestion queries, which require them.

PR #97, opened 2021-05-12
Fix filter suggests to work for views

fbertsch:
Fixes https://github.com/mozilla/lookml-generator/issues/90.

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -147,7 +147,7 @@ function generate_hub_commit() {
 
   HUB_DIR="looker-hub"
   GENERATION_DISALLOW_LIST="/app/lookml-generator/GENERATION_DISALLOW_LIST"
-  NAMESPACE_ALLOWLIST="/app/lookml-generator/namespace_allowlist.yaml"
+  NAMESPACE_ALLOWLIST="/app/lookml-generator/namespaces-allowlist.yaml"
   CUSTOM_NAMESPACES_FILENAME="/app/lookml-generator/custom-namespaces.yaml"
   GENERATED_SQL_URI="https://github.com/mozilla/bigquery-etl/archive/generated-sql.tar.gz"
   APP_LISTINGS_URI="https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings"
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -25,6 +25,19 @@ class Explore:
 
     def to_lookml(self) -> dict:
         """Generate LookML for this explore."""
+        base_lookml = {}
+        for view_type, view in self.views.items():
+            if "join" in view_type:
+                continue
+            if self._get_view_has_submission(view):
+                base_lookml[
+                    "sql_always_where"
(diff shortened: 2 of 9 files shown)


generator/explores/explore.py
```diff
@@ -25,6 +25,19 @@ def to_dict(self) -> dict:
 
     def to_lookml(self) -> dict:
         """Generate LookML for this explore."""
+        base_lookml = {}
+        for view_type, view in self.views.items():
+            if "join" in view_type:
```
wlach, line 30: Can you add some comments around here?

generator/lkml_update.py
```diff
@@ -49,4 +48,5 @@ def parse_any(
                 name = value.pop("name")
             return self.parse_block(key, value, name)
         else:
+            print(f"Accessing key {key!r}")
```
wlach, line 51: Stray debugging line?

```suggestion
```

bin/generate
```diff
@@ -147,7 +147,7 @@ function generate_hub_commit() {
 
   HUB_DIR="looker-hub"
   GENERATION_DISALLOW_LIST="/app/lookml-generator/GENERATION_DISALLOW_LIST"
-  NAMESPACE_ALLOWLIST="/app/lookml-generator/namespace_allowlist.yaml"
+  NAMESPACE_ALLOWLIST="/app/lookml-generator/namespaces-allowlist.yaml"
```
wlach, line 150: I didn't see the PR go by where this changed? I'm sure it's fine.

generator/explores/explore.py
```diff
@@ -25,6 +25,19 @@ def to_dict(self) -> dict:
...
+                continue
+            if self._get_view_has_submission(view):
+                base_lookml[
+                    "sql_always_where"
+                ] = f"${{{self.name}.submission_date}} >= '2010-01-01'"
+
+        base_lookml.update(self._to_lookml())
```
wlach, line 37: Is there a reason why you're updating this here, and not just saying `base_lookml = self._to_lookml()` on top? Do you want `sql_always_where` to be overriden by the definition? If so I'd try to make this explicit, maybe with some comments (as suggested just above)

wlach (approved):
No major concerns, r+wc

generator/explores/explore.py
fbertsch: I will - I think we need to formalize the `view_types` a bit more. The way I'm starting to see it in my head is:
`base_view` - That's the one this explore is created from. Usually the explore name will match this view's name
`extended_view*` - Any views we are extending to create the base view. For example, `client_counts` extends `baseline_clients_last_seen_table`
`join_view*` - Any views we are joining in the explore. This will be for e.g. event funnels

bin/generate
fbertsch: That was here: https://github.com/mozilla/lookml-generator/pull/93

generator/explores/explore.py
fbertsch: Indeed, the `_to_lookml` should take precedence. I'll add a comment

PR #102, opened 2021-05-12
Enable bergamot

fbertsch:
Tested this on dev, a few things are odd (e.g. test metrics), but we can fix those up later.

### DIFF
--- a/namespaces-allowlist.yaml
+++ b/namespaces-allowlist.yaml
@@ -6,3 +6,4 @@ firefox_ios:
 fenix:
   owners:
     - vzare@mozilla.com
+bergamot: null


namespaces-allowlist.yaml
```diff
@@ -6,3 +6,4 @@ firefox_ios:
 fenix:
   owners:
     - vzare@mozilla.com
+bergamot: null
```
wlach, line 9: Should this have `owners`?

wlach (approved):

namespaces-allowlist.yaml
fbertsch: `owners` is optional, and simply appends additional owners to the ones from the probe-info-service. If they need more we can add them later.

PR #108, opened 2021-05-18
Tighter version pinning of docker containers, update python requirements

wlach:
~~It appears that something broke with 3.8.10?~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN apt-add-repository https://cli.github.com/packages
 RUN apt update
 RUN apt install -y gh
 
-COPY --from=google/cloud-sdk:alpine /google-cloud-sdk /google-cloud-sdk
+COPY --from=google/cloud-sdk:339.0.0-alpine /google-cloud-sdk /google-cloud-sdk
 ENV PATH /google-cloud-sdk/bin:$PATH
 
 WORKDIR ${HOME}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8-slim
+FROM python:3.8.9-slim
 MAINTAINER Frank Bertsch <frank@mozilla.com>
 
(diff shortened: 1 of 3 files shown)


wlach:
The pinning I tried earlier doesn't do any good, but I wonder if we should keep it to rule out dependency changes as a source of bugs in the futre.

wlach:
Upgrading the dependencies seems to fix the problems we're seeing. I believe there's an upstream issue where dependencies are being incorrectly calculated-- It seems like a specific dependency (`google-cloud-core`) is asking for the latest version of a dep (`google-api-core`) even though a compatible version is already installed. 

```
Collecting google-api-core[grpc]==1.26.3
  Downloading google_api_core-1.26.3-py2.py3-none-any.whl (93 kB)
... (more stuff happens) ...
Collecting google-api-core<2.0.0dev,>=1.21.0
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    google-api-core<2.0.0dev,>=1.21.0 from https://files.pythonhosted.org/packages/ae/27/a6d8f72a2d2cc98d5b5ccd05fbfdbe99f07625cc9e51727b5f8e6a53c470/google_api_core-1.27.0-py2.py3-none-any.whl#sha256=edfca00d76081f359bed7b2efa2dc3a057309e55bf0a91628a0639730e31c357 (from google-cloud-core==1.6.0->-r /app/requirements.txt (line 132))
```

Upgrading that dependency to the latest fixes the issue (I also removed it from `requirements.in`, since that doesn't do any good). Short of fixing the bug upstream, one possible way of making this easier in the future would be to enable dependabot updates for this repo. In this case that would have saved a lot of hand-wringing (since we would see the dependency bump that fixed the issue in our PR list)

relud (approved):

fbertsch (approved):
LGTM

PR #109, opened 2021-05-18
Add dependabot configuration

wlach:


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+version: 2
+updates:
+  - package-ecosystem: pip
+    directory: /
+    schedule:
+      interval: daily
+    reviewers:
+      - relud
+      - fbertsch
+      - wlach
```
relud, line 11: @fbertsch how would you feel about using a group review request here, as seen for https://github.com/mozilla/looker-spoke-default/pull/49

relud (approved):
r+, comment can fixed in a separate PR

.github/dependabot.yml
fbertsch: Big fan of the group reviews. That would be a nice improvement

PR #121, opened 2021-05-20
Bug 1707727 - Upload namespaces.yaml to a GCS bucket

wlach:
This will let us consume the published namespaces.yaml from the
Glean Dictionary (see https://github.com/mozilla/glean-dictionary/pull/603)

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -151,6 +151,7 @@ function generate_hub_commit() {
   CUSTOM_NAMESPACES_FILENAME="/app/lookml-generator/custom-namespaces.yaml"
   GENERATED_SQL_URI="https://github.com/mozilla/bigquery-etl/archive/generated-sql.tar.gz"
   APP_LISTINGS_URI="https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings"
+  NAMESPACE_GCS_BUCKET="wlach-test-gcs"
 
   # Generate namespaces.yaml and LookML
   lookml-generator namespaces \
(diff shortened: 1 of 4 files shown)


bin/generate
```diff
@@ -151,13 +151,15 @@ function generate_hub_commit() {
   CUSTOM_NAMESPACES_FILENAME="/app/lookml-generator/custom-namespaces.yaml"
   GENERATED_SQL_URI="https://github.com/mozilla/bigquery-etl/archive/generated-sql.tar.gz"
   APP_LISTINGS_URI="https://probeinfo.telemetry.mozilla.org/v2/glean/app-listings"
+  NAMESPACE_GCS_BUCKET="wlach-test-gcs"
```
wlach, line 154: This will obviously need to be changed before we land. @haroldwoo I guess it would be up to you to set this bucket up and make sure the container has write permissions to it (and that we can read from it via an unauthenticated https session).

wlach:
Have tested this locally and it works fine.

whd:
To make this artifact more like `generated-sql` and `generated-schemas` I propose we push this to a git repository instead of a GCS bucket. From discussion with :frank it sounds like https://github.com/mozilla/looker-hub is already used for generated looker files and putting it at the top level of the default branch of that repo would be appropriate. It would be accessible over HTTPS via github and we already have a machine user allocated that we can provide credentials to for whatever process ends up being responsible for running `lookml-generator`.

wlach:
Thanks @whd I'll abandon this PR and start a new one

PR #131, opened 2021-05-25
Funnel analysis explores

### LINKED ISSUE #76: Generate funnel analysis explores
The [funnel analysis view](https://github.com/mozilla/looker-spoke-default/tree/main/user_journey/views/funnel_analysis) in UJET is a good basis for all funnel analyses explores. The initial work here will not include event_properties, but we can expand and include them later.

### LINKED ISSUE #87: Separate tests for different view and explore types
We currently test a majority of the code in the same test function: https://github.com/mozilla/lookml-generator/blob/main/tests/test_lookml.py. The `test_lookml_actual` is run on a variety of explores and views, and the test fixtures contain the data for all of the different types.

Instead, we should separate those out so each one is tested separately. This would solve two problems:
1. Make it easier to understand failures. If there's an issue during lookml generation, it's not always clear which kind of view or explore the failure happened in, and requires logging/debugging to find out.
2. Make it easier to add new views and explores. With a common pattern of adding new views and explores, it should be less overhead for developers to add news ones. For example, we should contain all the associated test fixtures (where it makes sense) in the associated test.

fbertsch:
Fixes #76 and Fixes #87. New views/explores should be able to follow this pattern of "Create a test file".

Tested on the branch `funnel_analysis`, there will be an explore on that branch for Fenix funnels.

### DIFF
--- a/generator/views/funnel_analysis_view.py
+++ b/generator/views/funnel_analysis_view.py
@@ -126,8 +126,8 @@ class FunnelAnalysisView(View):
                             "FROM "
                             f"{self.tables[0]['event_types']} "
                             "WHERE "
-                            "{% condition message_id %} event_types.category {% endcondition %} "
-                            "AND {% condition event_type %} event_types.event {% endcondition %}"
+                            "{% condition category %} category {% endcondition %} "
+                            "AND {% condition event %} event {% endcondition %}"
                         )
                     },
                     "filters": [
@@ -138,10 +138,10 @@ class FunnelAnalysisView(View):
                             "suggest_dimension": "event_names.category",
                         },
                         {
-                            "name": "name",
+                            "name": "event",
                             "type": "string",
                             "suggest_explore": "event_names",
-                            "suggest_dimension": "event_names.name",
+                            "suggest_dimension": "event_names.event",
                         },
                     ],
                     "dimensions": [
--- a/generator/views/funnel_analysis_view.py
+++ b/generator/views/funnel_analysis_view.py
@@ -1,0 +1,48 @@
+"""Class to describe a Funnel Analysis View."""
+from __future__ import annotations
+
+from typing import Dict, Iterator, List
+
+from .view import View
+
+
+class FunnelAnalysisView(View):
+    """A view for Client Counting measures."""
+
+    type: str = "funnel_analysis_view"
+
+    def __init__(self, namespace: str, tables: List[Dict[str, str]]):
+        """Get an instance of a FunnelAnalysisView."""
+        super().__init__(namespace, "funnel_analysis", FunnelAnalysisView.type, tables)
+
+    @classmethod
+    def from_db_views(
+        klass,
+        namespace: str,
+        is_glean: bool,
+        channels: List[Dict[str, str]],
+        db_views: dict,
+    ) -> Iterator[FunnelAnalysisView]:
+        """Get Client Count Views from db views and app variants."""
+        # We can guarantee there will always be at least one channel,
+        # because this comes from the associated _get_glean_repos in
+        # namespaces.py
+        dataset = next(
+            (channel for channel in channels if channel.get("channel") == "release"),
+            channels[0],
+        )["dataset"]
+
+        necessary_views = {"events_daily", "event_types"}
+        for view_id, references in db_views[dataset].items():
+            necessary_views -= {view_id}
+
+        if len(necessary_views) == 0:
+            yield FunnelAnalysisView(
+                namespace,
+                [
+                    {
+                        "events_daily_view": "events_daily_table",
+                        "event_types_view": "event_types_table",
+                    },
+                ],
+            )
--- a/generator/views/funnel_analysis_view.py
+++ b/generator/views/funnel_analysis_view.py
@@ -113,50 +113,91 @@ class FunnelAnalysisView(View):
         ]
 
     def _event_types_lookml(self) -> List[Dict[str, Any]]:
-        events = [
-            {
-                "name": "event_types",
-                "derived_table": {
-                    "sql": (
-                        "SELECT "
-                        "mozfun.event_analysis.aggregate_match_strings( "
-                        "ARRAY_AGG( "
-                        "mozfun.event_analysis.event_index_to_match_string(index))) AS match_string "
-                        "FROM "
-                        f"{self.tables[0]['event_types']} "
-                        "WHERE "
-                        "{% condition message_id %} event_types.category {% endcondition %} "
-                        "AND {% condition event_type %} event_types.event {% endcondition %}"
-                    )
-                },
-                "filters": [
-                    {
-                        "name": "category",
-                        "type": "string",
-                        "suggest_explore": "event_names",
-                        "suggest_dimension": "event_names.category",
+        events = (
+            [
+                {
+                    "name": "event_types",
+                    "derived_table": {
+                        "sql": (
+                            "SELECT "
+                            "mozfun.event_analysis.aggregate_match_strings( "
+                            "ARRAY_AGG( "
+                            "mozfun.event_analysis.event_index_to_match_string(index))) AS match_string "
+                            "FROM "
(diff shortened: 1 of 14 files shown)


generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,203 @@
...
+from typing import Any, Dict, Iterator, List, Optional
+
+from .view import View, ViewDict
+
+
+class FunnelAnalysisView(View):
+    """A view for Client Counting measures."""
+
+    type: str = "funnel_analysis_view"
+    num_funnel_steps: int = 4
```
wlach, line 13:
```suggestion
        DEFAULT_NUM_FUNNEL_STEPS: int = 4
```

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,203 @@
...
+            channels[0],
+        )["dataset"]
+
+        necessary_views = {"events_daily", "event_types"}
+        actual_views = {}
+        for view_id, references in db_views[dataset].items():
+            if view_id in necessary_views:
+                actual_views[view_id] = f"`mozdata.{dataset}.{view_id}`"
+
+        if len(actual_views) == 2:
```
wlach, line 43: So if this isn't true it yields nothing? Some comments to explain the behaviour here might be good.

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,203 @@
...
+                            "SELECT category, "
+                            "  event, "
+                            "  property.key AS property_name, "
+                            "  property_value.key AS property_value, "
+                            f"FROM {self.tables[0]['event_types']} "
+                            "LEFT JOIN UNNEST(event_properties) AS property "
+                            "LEFT JOIN UNNEST(property.value) AS property_value "
+                        )
+                    },
+                    "dimensions": [
```
wlach, line 177: Can you add descriptions for these?

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,203 @@
+"""Class to describe a Funnel Analysis View."""
```
wlach, line 1: Can you add a few more comments about what this is and the reasoning behind its architecture?

wlach (commented):
Quite a bit to unpack here-- I left some early comments. I wouldn't mind a bit more comments/context to help interpret the view. I'm a bit worried I'm reviewing something I don't fully understand and I think the extra documentation might help someone who stumbles across this in the future.

generator/views/funnel_analysis_view.py
fbertsch: Right, we only yield a funnel analysis view if we have all the underlying tables. Will add a comment.

generator/views/funnel_analysis_view.py
fbertsch: Good call!

fbertsch:
@wlach I'm hoping that between these updated comments and the new doc for `namespaces.yaml`, this should be clearer.

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,238 @@
...
+                            f"{self.tables[0]['event_types']} "
+                            "WHERE "
+                            "{% condition category %} category {% endcondition %} "
+                            "AND {% condition event %} event {% endcondition %}"
+                        )
+                    },
+                    "filters": [
+                        {
+                            "name": "category",
+                            "description": "The event category, as defined in metrics.yaml.",
```
wlach, line 169: Let's go for something more human centred here. I'm assuming you mean metric category as defined inside Glean?

```suggestion
                            "description": "The Glean metric category, as defined inside the application.",
```

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,238 @@
...
+                            "type": "string",
+                            "suggest_explore": "event_names",
+                            "suggest_dimension": "event_names.category",
+                        },
+                        {
+                            "name": "event",
+                            "description": "The event name.",
```
wlach, line 176:
```suggestion
                            "description": "The Glean metric name, as defined inside the application.",
```

generator/views/funnel_analysis_view.py
wlach: I think these still need to be added :)

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,238 @@
+"""Class to describe a Funnel Analysis View.
```
wlach, line 1: This is great. I think it might be worth adding one or two sentences on how this view relates to the underlying dataset (with a link): https://docs.telemetry.mozilla.org/datasets/non_desktop/events_daily/reference.html

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,238 @@
...
+                ),
+            }
+            for n in range(1, self.n_events() + 1)
+        ]
+        count_measures: List[Dict[str, Any]] = [
+            {
+                "name": f"count_user_days_event_{n}",
+                "description": (
+                    f"The number of user-days that completed step {n}. "
+                    "Grouping by day makes this is a count of users."
```
wlach, line 123: This sentence doesn't quite parse for me.

generator/views/funnel_analysis_view.py
```diff
@@ -0,0 +1,238 @@
...
+        return {
+            "includes": [f"{self.tables[0]['funnel_analysis']}.view.lkml"],
+            "views": self._funnel_analysis_lookml() + self._event_types_lookml(),
+        }
+
+    def n_events(self) -> int:
+        """Get the number of events allowed in this funnel."""
+        return len([k for k in self.tables[0] if k.startswith("event_type_")])
+
+    def _funnel_analysis_lookml(self) -> List[Dict[str, Any]]:
```
wlach, line 104: This seems like the core of this explore. Taking it on faith here that this is debugged and works well :)

wlach (approved):
I worry a bit that I don't understand these funnels work well enough to give you a proper review here, but things look pretty good based on the understanding I do have. I mostly have some small suggestions around naming, etc.

generator/views/funnel_analysis_view.py
fbertsch: Ah, this view doesn't need descriptions. It's only used for filter suggestions.

generator/views/funnel_analysis_view.py
fbertsch: Okay, let me describe what this is and maybe you can help me uncover a better description.

For this explore, the analyst creates a funnel (step 1 -> step 2 -> ...). They then select this measure (e.g. `count_user_days_event_2`).

The underlying table has a grain of client-day: so for each client-day, we see if that funnel was completed. So if you simply selected this value, it would be the number of client-days that completed the funnel; however if you then selected `submission_date` as a dimension, each `submission_date` would give the number of clients who completed the funnel.

generator/views/funnel_analysis_view.py
fbertsch: Yup, this is based on the [UJET funnel analysis explore](https://github.com/mozilla/looker-spoke-default/blob/main/user_journey/views/funnel_analysis/funnel_analysis.view.lkml). That has been sufficiently debugged and used.

generator/views/funnel_analysis_view.py
wlach: How about this (hoping I understand your meaning):

`The total number of times that step {n} was completed. Grouping by day makes this a count of users who completed step {n} on each day.`

I would avoid using bespoke terminology like "client day" if at all possible.

fbertsch: I like that, but then I have no idea what to call the measure. Maybe `count_completed_event_{n}`?

wlach: I'm less worried about the name than the description but sounds good!

PR #139, opened 2021-05-27
Add unnested labeled counter to glean ping explores

acmiyaguchi:
wip.

I'm in the process of figuring out how to add view joins to the main glean ping explores.


acmiyaguchi:
Working through this, the relationship between the View, PingView, and GleanPingView have been a little confusing to work though. I initially thought I would start with a new class as an `GleanUnnestedLabelView` which would generate it's own lookml definitions. I may still do this, since a suggest explore will need to be made for each join view that's included in the explore. At this point, it adds a lot of extra complexity.

acmiyaguchi:
@fbertsch do you have suggestions on how to proceed? Basically I'd like to extend this definition in the GleanPingExplore:

https://github.com/mozilla/lookml-generator/blob/53761647814749cb66deef0574d3c4f78ecf534a/generator/explores/ping_explore.py#L16-L24

However, it doesn't look like the explore has the full context of the view definition. If I had the view defintions, I would do something like this:

```python
# to get the name of the first explore
base = view["views"][0]

# sql along the lines of
# CROSS JOIN UNNEST(${{{base_view}.{metric_name}}}) as ${{{unnested_view}}} ;; 

joins = [{"relationship": "one_to_many", "sql": ...} for definition in view["views"][1:]]
```

This would let me get the name of the view for doing the cross join. Any pointers on where to look?

fbertsch:
The other option here is finding the associated view definition from `namespaces.yaml`, or the actual view instance itself, and pushing it into the `get_lookml` function. That might make sense from a flow perspective.

acmiyaguchi:
Aha, the `get_view_lookml` method was exactly what I was looking for.

fbertsch:
@acmiyaguchi I just realized - are you enabling these views across all nested fields? That's probably how this should go, and then `glean_ping_view` can add just those measures to the nested views.

fbertsch:
@acmiyaguchi It's probably best to first think about including nested fields, then add the ability for the counter measures. It wouldn't be great to have separate implementations for different repeated fields.

acmiyaguchi:
I'd like to complete this train of thought to the end, I think it'll be easier to refactor this behavior once there are concrete tests than to work from scratch. I have a clear idea of what the nested views should look like labeled counters (and to a lesser degree, most of the nested fields). 

Generalizing this is more difficult because of the single view assumptions that
 `PingView::get_dimensions` and `PingView::get_measures` are making, which is polymorphically called in the `TableView::to_lookml` method. Getting the dimensions and measures for the join views (which I think belong inside the same file as the base view) means that there needs to be support for iterating over all the dimensions and grouping them based on which view it's part of.

If I inherit from the GleanPingView to make a new object per nested view, I would also need to have a new view file per object because I wouldn't be able to call the derived class within `GleanPingView::to_lookml` because of circular dependencies. This was the approach that I was taking (captured in the commit history), and it was just all too wildly complex while trying to grok the code base.

fbertsch:
Okay, I understand that the complication is great - it sounds like you've thought about the implementation, so this should be at least roughly forwards compatible. I appreciate your thoughtful response.

acmiyaguchi:
This PR is functional -- I'm able to query labeled counters. I'll be adding in suggest explores/views which will require modifying the public interface for `Explore::to_lookml` to return an iterable of explores vs a single explore.

@fbertsch any feedback at this point on the PR?

acmiyaguchi:
https://github.com/mozilla/looker-hub/compare/amiyaguchi-generation-test-1

Generated diff for reference.

generator/explores/glean_ping_explore.py
```diff
@@ -10,9 +11,43 @@ class GleanPingExplore(PingExplore):
...
     def from_views(views: List[View]) -> Iterator[PingExplore]:
         """Generate all possible GleanPingExplores from the views."""
         for view in views:
             if view.view_type == GleanPingView.type:
                 yield GleanPingExplore(view.name, {"base_view": view.name})
+
+    @staticmethod
+    def from_dict(name: str, defn: dict, views_path: Path) -> PingExplore:
+        """Get an instance of this explore from a name and dictionary definition."""
+        return GleanPingExplore(name, defn["views"], views_path)
```
acmiyaguchi, line 53: This method wasn't implemented, which was causing it to return an instance of PingExplore instead. Took a while to debug.

generator/views/glean_ping_view.py
```diff
@@ -35,6 +36,66 @@ class GleanPingView(PingView):
...
+            (table for table in self.tables if table.get("channel") == "release"),
+            self.tables[0],
+        )["table"]
+        dimensions = self.get_dimensions(bq_client, table, v1_name)
+
+        client_id_field = self._get_client_id(dimensions, table)
+
+        view_definitions = []
+        metrics = self._get_glean_metrics(v1_name)
+        for metric in metrics:
```
acmiyaguchi, line 60: Perhaps this should iterate over dimensions instead of the metrics.

generator/views/glean_ping_view.py
```diff
@@ -35,6 +36,66 @@ class GleanPingView(PingView):
...
+                        {
+                            "name": "client_count",
+                            "type": "count_distinct",
+                            "sql": f"case when ${{value}} > 0 then ${{{self.name}.{client_id_field}}} end",
+                        },
+                    ],
+                }
+                view_definitions += [definition]
+        # deduplicate view definitions, because somehow a few entries make it in
+        # twice e.g. metrics__metrics__labeled_counter__media_audio_init_failure
```
acmiyaguchi, line 92: Do you have a sense for why there are duplicate defintions? Is it because there are sub-dimensions for the metric?

tests/test_lookml.py
```diff
@@ -560,6 +580,16 @@ def test_lookml_actual(mock_glean_ping, runner, glean_apps, tmp_path, msg_glean_
...
+        yield namespaces_text
+
+
+@patch("generator.views.glean_ping_view.GleanPing")
+def test_lookml_actual_baseline_view(
+    mock_glean_ping, runner, glean_apps, tmp_path, msg_glean_probes
+):
+    with _prepare_lookml_actual_test(
+        mock_glean_ping, runner, glean_apps, tmp_path, msg_glean_probes
+    ) as namespaces_text:
```
acmiyaguchi, line 592: Refactored the tests so there is a single test per unit instead of one large test. This could be refactored further to reduce boilerplate, but leaving it at this indentation has a benefit of making the diff relatively small.

generator/explores/glean_ping_explore.py
```diff
@@ -10,9 +11,43 @@ class GleanPingExplore(PingExplore):
...
+        base_name = base["name"]
+
+        joins = []
+        for view in views_lookml["views"][1:]:
+            view_name = view["name"]
+            metric = "__".join(view["name"].split("__")[1:])
+            join = {
+                "name": view["name"],
+                "relationship": "one_to_many",
+                "sql": f"CROSS JOIN UNNEST(${{{base_name}.{metric}}}) AS {view_name}",
```
fbertsch, line 30: Looking over the explore, we should probably use a `LEFT JOIN UNNEST`. That would allow users to get client counts for many different counters without affecting the underlying table (i.e. all clients would still be present).

generator/views/glean_ping_view.py
```diff
@@ -83,20 +144,30 @@ def _get_glean_metrics(self, v1_name: Optional[str]) -> List[GleanProbe]:
...
+        sep = "" if not category else "_"
         label = name
-        sep = "_"
-        if not category:
-            sep = ""
         looker_name = f"metrics__{metric.type}__{category}{sep}{name}"
         if suffix:
             label = f"{name}_{suffix}"
-            looker_name = f"metrics__{metric.type}__{category}{sep}{name}__{suffix}"
+            looker_name = f"{looker_name}__{suffix}"
```
fbertsch, line 170: thanks for cleaning these up!

fbertsch (commented):
The approach looks good, the test refactor is an improvement 👍

acmiyaguchi:
Suggest explores are working as expected, but there are some unit tests that I need to fix before marking this ready.

generator/explores/glean_ping_explore.py
acmiyaguchi: Is left join unnest possible without having a primary key?

acmiyaguchi: I played with this for a bit; it seems like choosing a LEFT JOIN will have many of the same constraints as the CROSS JOIN, except that it also includes the null value as a field. Should each view hold a reference to the client id, so we can join on it as the primary key?

acmiyaguchi: I set the primary key to the document id, used a left join. With this, it's possible to join between the different unnested views. However, there are issues using sums from the base table with the unnested one, results in:

```
A LookML model issue occurred.
Non-unique value/primary key (or sql_distinct_key), value overflow or collision when computing sum.
```

The cross join did not have the ability to compare labeled measures against each other. The left join alone had issues computing the sum of aggregates for a single unnested table.

generator/explores/glean_ping_explore.py
```diff
@@ -26,12 +26,48 @@ def _to_lookml(self, v1_name: Optional[str]) -> List[Dict[str, Any]]:
...
+                continue
+            view_name = view["name"]
+            metric = "__".join(view["name"].split("__")[1:])
+            joins.append(
+                {
+                    "name": view_name,
+                    "relationship": "one_to_many",
+                    "sql": (
+                        f"LEFT JOIN UNNEST(${{{base_name}.{metric}}}) AS {view_name} "
+                        f"ON ${{{base_name}.document_id}} = ${{{view_name}.document_id}}"
```
acmiyaguchi, line 49: Not every table has a document_id field, so this wouldn't work generically.

acmiyaguchi (commented):
This is ready for review, after rebasing and testing with the fenix dataset.

generator/views/ping_view.py
```diff
@@ -69,8 +69,16 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
...
+        # set document id field as a primary key for joins
         view_defn["dimensions"] = list(
-            filterfalse(lookml_utils._is_dimension_group, dimensions)
+            filterfalse(
+                lookml_utils._is_dimension_group,
+                [
+                    d if d["name"] != "document_id" else dict(**d, primary_key="yes")
+                    for d in dimensions
+                ],
+            )
```
fbertsch, line 81:
```suggestion
           [
                d if d["name"] != "document_id" else dict(**d, primary_key="yes")
                for d in dimensions
                if not lookml_utils._is_dimension_group(d)
            ]
```

tests/test_lookml.py
```diff
@@ -597,6 +614,27 @@ def test_lookml_actual(
...
+@patch("generator.explores.glean_ping_explore.GleanPing")
+def test_lookml_actual_baseline_view(
+    mock_glean_ping_view,
+    mock_glean_ping_explore,
+    runner,
+    glean_apps,
+    tmp_path,
+    msg_glean_probes,
+):
+    with _prepare_lookml_actual_test(
```
fbertsch, line 630: _so_ much better

tests/test_lookml.py
```diff
@@ -1123,13 +1202,88 @@ def test_lookml_actual(
...
                         },
                     ],
-                }
+                },
+                {
+                    "dimensions": [
+                        {
+                            "hidden": "yes",
+                            "name": "document_id",
+                            "primary_key": "yes",
```
fbertsch, line 1211: `document_id` is not a primary key for this table, right? e.g.
```
SELECT document_id, label, val
FROM mozdata.org_mozilla_firefox.metrics
LEFT JOIN UNNEST(metrics.labeled_counters.some_labeled_counter)
```

Will result in:
```
document_id | label | value
abc | label1 | val1
abc | label2 | val2
abc | label3 | val3
```
So the docid does not uniquely identify a row.

fbertsch (commented):
Everything looks good except the doc id bits. Have you pushed this to a branch on `looker-hub` I can test out?

acmiyaguchi:
> Everything looks good except the doc id bits. Have you pushed this to a branch on looker-hub I can test out?

What's currently in this PR should be pushed to `amiyaguchi-generation-test-1` (https://github.com/mozilla/looker-hub/compare/amiyaguchi-generation-test-1)

I probably should have pushed to different branches for the various join conditions, but i just rewrote the same branch.

generator/views/glean_ping_view.py
```diff
@@ -35,6 +37,103 @@ class GleanPingView(PingView):
...
+                    "name": suggest_name,
+                    "derived_table": {
+                        "sql": dedent(
+                            f"""
+                            select
+                                m.key,
+                                count(*) as n
+                            from {table} as t,
+                            unnest(metrics.{metric.type}.{metric.id.replace(".", "_")}) as m
+                            where date(submission_timestamp) > date_sub(current_date, interval 3 day)
```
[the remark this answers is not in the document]
acmiyaguchi: It needs a date range filter, but maybe a combination of 3 months + 1% sample would work well.

tests/test_lookml.py
acmiyaguchi: I don't think there's actually a primary key anywhere, but the join condition will complain for anything that isn't a cross join.

generator/explores/glean_ping_explore.py
acmiyaguchi: If I understand correctly, yes. Without a join key, there are issues with counting across different labeled counters.

generator/explores/glean_ping_explore.py
```diff
@@ -26,12 +26,48 @@ def _to_lookml(self, v1_name: Optional[str]) -> List[Dict[str, Any]]:
...
-        # insert the description in
-        lookml = super()._to_lookml(v1_name)
-        lookml[0][
-            "description"
-        ] = f"Explore for the {self.name} ping. {ping_description}"
-        return lookml
+
+        views_lookml = self.get_view_lookml(self.views["base_view"])
+
+        # The first view, by convention, is always the base view with the
```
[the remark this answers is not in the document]
acmiyaguchi: I have a PR for the docs I want to add, I'll make sure to add it there.

generator/explores/glean_ping_explore.py
fbertsch: We should maybe drop the primary key completely. See below, I tried to use it and it fails. You should be fine to use the LEFT JOIN without a join key.

Screen Shot 2021-06-03 at 3 34 33 PM

tests/test_lookml.py
fbertsch: Can you show what you tried without a JOIN key and what the error was?

acmiyaguchi:
I created 3 new branches to test out the different join behaviors:

* `amiyaguchi-generation-left-join-doc` - left join on document id
* amiyaguchi-generation-left-join-no-key` - left join no primary key
* amiyaguchi-generation-cross-join-no-key` - cross join no primary key

Testing procedure:

- Create explore with 7 day filter
- Add labeled counter > metrics search counts > key, client count, and count
- Add labeled counter > crash metrics crash > key, client count, and count
- Add metrics > metrics > tabs open count


* join-left-doc: query works fine
* join-left-no-key
    -`A LookML model issue occurred
Field 'Metrics Metrics Tabs Open Count' cannot be calculated because of a one_to_many or many_to_many join. Add a primary key to metrics.
spoke-default/imported_projects/looker-hub/fenix/views/metrics.view.lkml:4589`
    - removing tabs open count allows query to complete, result set is >500 rows
    - removing crash metrics crash and adding tabs open count leads to original issue
* cross-join-no-key
    - default result on two unnested tables are missing null values are included and result set has ~300 rows. Values look roughly correct wrt to left join
    - adding tab counts causes issue: `A LookML model issue occurred.
Field 'Metrics Metrics Tabs Open Count' cannot be calculated because of a one_to_many or many_to_many join. Add a primary key to metrics.
spoke-default/imported_projects/looker-hub/fenix/views/metrics.view.lkml:4589`
    - removing crash metrics crash leads to same issue

So without the primary key, it's impossible to look at metrics in the unnested table with the metrics the base table. I'm not entirely sure if using the document id as a primary key is the correct thing to do if there are duplicates across date boundaries, but it is the only option out of the ones that I've tested that allow for aggregates across labeled and unlabeled counters. That might not actually be an issue though, since dimensions from the base table seem to work correctly (and may suffice for most explores).

@fbertsch your opinion on which option to go with here? The left join is definitely better than the cross join, but whether to use a primary key seems to be the question here. In a more general sense, it would be possible to add document ids for ping tables, but determining primary keys of the derived tables is not so easy.

fbertsch:
@acmiyaguchi I think it's worth adding the primary key to the base tables. For derived tables, we can handle it on a case-by-case basis; and things should still work even without a primary key (just not as well). For most of them, though, it will probably be `CONCAT(client_id, date)`. Then we'll use the LEFT JOIN, and not have a primary key on the joined tables.

tests/test_lookml.py
acmiyaguchi: I think I understand this better, it is the primary key for a query like this:

```
with labeled_counter as (
  SELECT document_id, label, val
  FROM mozdata.org_mozilla_firefox.metrics t
  LEFT JOIN UNNEST(metrics.labeled_counters.some_labeled_counter)
),
counter as (
  SELECT document_id, metrics.counter.some_counter
  FROM mozdata.org_mozilla_firefox.metrics t
)
select document_id, counter, label, val
from counter
left join labeled_counter
using (document_id)
```

Which i think is the conceptual join that's happening here, with the symmetric aggregation magic that looker is doing under the hood.

generator/views/glean_ping_view.py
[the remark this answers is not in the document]
acmiyaguchi: I'm going to reduce this to a month, since the query time is noticeable.

acmiyaguchi:
I'm comfortable merging with this iteration for labeled counters. I'll enable auto-merge. There are build issues unrelated to this PR though.

fbertsch (approved):
Okay, I think we're ready to test this on stage.

PR #142, opened 2021-06-01
Add remaining contextual-services tables

fbertsch:
I figured we should enable all the contextual services data and hold a demo at some point with the team. @BenWu I was under the impression that you've been working with this, right?


BenWu (approved):
Looks consistent with the rest of the file

PR #143, opened 2021-06-01
Generate descriptions for Glean Ping explores (fixes #106)

### LINKED ISSUE #106: Include Glean Ping descriptions in generated LookML
Looker Explores have an [optional description field](https://docs.looker.com/reference/explore-params/description-for-explore) that displays as a tooltip next to the explore in the drop-down. For Glean Ping explores, we should enable that tooltip by pulling in the description through the probe-info-service [/pings endpoint](https://probeinfo.telemetry.mozilla.org/glean/fenix/pings). For example, the activation ping for Fenix explore would have the description `This ping is intended to provide a measure of the activation of mobile\nproducts. It's generated when Fenix starts, right after Glean is\ninitialized. It doesn't include the client_id, since it might be reporting\nan hashed version of the Google Advertising ID.`.

wlach:
Format should be `Explore for <x> ping. <x ping description>`. For
example, for the metrics ping, you would see:

```
Explore for the metrics ping. The metrics ping is intended for all of
the metrics that are explicitly set by the application or are included
in the application's metrics.yaml file (except events). The reported
data is tied to the ping's measurement window, which is the time
between the collection of two metrics ping. Ideally, this window is
expected to be about 24 hours, given that the collection is scheduled
daily at 4AM. Data in the ping_info section of the ping can be used to
infer the length of this window.
```

### DIFF
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -15,6 +15,7 @@ class Explore:
     """A generic explore."""
 
     name: str
+    namespace: str
     views: Dict[str, str]
     views_path: Optional[Path] = None
     type: str = field(init=False)
@@ -23,7 +24,7 @@ class Explore:
         """Explore instance represented as a dict."""
         return {self.name: {"type": self.type, "views": self.views}}
 
-    def to_lookml(self) -> List[Dict[str, Any]]:
+    def to_lookml(self, v1_name: Optional[str]) -> List[Dict[str, Any]]:
         """
         Generate LookML for this explore.
 
@@ -58,13 +59,13 @@ class Explore:
                 ] = f"${{{base_view_name}.submission_date}} >= '2010-01-01'"
 
         # We only update the first returned explore
-        new_lookml = self._to_lookml()
+        new_lookml = self._to_lookml(v1_name)
         base_lookml.update(new_lookml[0])
         new_lookml[0] = base_lookml
 
         return new_lookml
 
-    def _to_lookml(self) -> List[Dict[str, Any]]:
+    def _to_lookml(self, v1_name: Optional[str]) -> List[Dict[str, Any]]:
         raise NotImplementedError("Only implemented in subclasses")
 
     def get_dependent_views(self) -> List[str]:
@@ -76,7 +77,7 @@ class Explore:
         ]
 
     @staticmethod
-    def from_dict(name: str, defn: dict, views_path: Path) -> Explore:
+    def from_dict(name: str, namespace: str, defn: dict, views_path: Path) -> Explore:
         """Get an instance of an explore from a namespace definition."""
         raise NotImplementedError("Only implemented in subclasses")
 
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -15,7 +15,6 @@ class Explore:
     """A generic explore."""
 
     name: str
-    namespace: str
     views: Dict[str, str]
     views_path: Optional[Path] = None
     type: str = field(init=False)
@@ -77,7 +76,7 @@ class Explore:
         ]
 
     @staticmethod
-    def from_dict(name: str, namespace: str, defn: dict, views_path: Path) -> Explore:
+    def from_dict(name: str, defn: dict, views_path: Path) -> Explore:
         """Get an instance of an explore from a namespace definition."""
         raise NotImplementedError("Only implemented in subclasses")
 
--- a/generator/explores/glean_ping_explore.py
+++ b/generator/explores/glean_ping_explore.py
@@ -24,7 +24,7 @@ class GleanPingExplore(PingExplore):
         ping_descriptions = {
             k.replace("-", "_"): v for k, v in glean_app.get_ping_descriptions().items()
         }
-        ping_description = ping_descriptions[self.name]
+        ping_description = " ".join(ping_descriptions[self.name].split())
         lookml = super()._to_lookml(v1_name)
         lookml[0][
             "description"
--- a/generator/explores/glean_ping_explore.py
+++ b/generator/explores/glean_ping_explore.py
@@ -24,7 +24,9 @@ class GleanPingExplore(PingExplore):
         ping_descriptions = {
             k.replace("-", "_"): v for k, v in glean_app.get_ping_descriptions().items()
         }
+        # collapse whitespace in the description so the lookml looks a little better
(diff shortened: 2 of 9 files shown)


tests/test_lookml.py
```diff
@@ -1137,6 +1155,7 @@ def test_lookml_actual(mock_glean_ping, runner, glean_apps, tmp_path, msg_glean_
             "explores": [
                 {
                     "name": "baseline",
+                    "description": "Explore for the baseline ping. The baseline ping is foo.",
```
wlach, line 1158: I didn't test the actual output, but this seems to indicate it should work.

wlach:
Was so excited to get this working that I forgot an important detail. This depends on https://github.com/mozilla/mozilla-schema-generator/pull/197 - need to merge that, do a new release of MSG, then update the dependencies in this PR.

generator/lookml.py
```diff
@@ -27,11 +27,20 @@ def _generate_views(
...
+    views_dir: Path,
+    v1_name: Optional[str],
 ) -> Iterable[Path]:
     for explore_name, defn in explores.items():
         logging.info(f"Generating lookml for explore {explore_name} in {namespace}")
-        explore = EXPLORE_TYPES[defn["type"]].from_dict(explore_name, defn, views_dir)
+        explore = EXPLORE_TYPES[defn["type"]].from_dict(
+            explore_name, namespace, defn, views_dir
+        )
+        print(explore.get_dependent_views())
```
fbertsch, line 42:
```suggestion
```

generator/lookml.py
```diff
@@ -27,11 +27,20 @@ def _generate_views(
 
 
 def _generate_explores(
-    client, out_dir: Path, namespace: str, explores: dict, views_dir: Path
+    client,
+    out_dir: Path,
+    namespace: str,
+    explores: dict,
+    views_dir: Path,
+    v1_name: Optional[str],
```
fbertsch, line 35: I regret now calling this `v1_name`. No need to change it everywhere (but feel free if you want to), however, can you add a doc line here about what it is?

tests/test_lookml.py
fbertsch: Okay, we should test this inside Looker. Run the following:
```
export HUB_BRANCH_PUBLISH="glean-ping-descriptions"
export SPOKE_BRANCH_PUBLISH="glean-ping-descriptions"
export GIT_SSH_KEY_BASE64=$(cat ~/.ssh/id_rsa | base64)
make build && make run
```

And it should push to `glean-ping-descriptions` in looker-hub and `glean-ping-descriptions-working` in looker-spoke-default. You can check those two branches out in Looker and ensure that the docs look correct and the LookML isn't broken.

generator/explores/glean_ping_explore.py
```diff
@@ -10,9 +15,28 @@ class GleanPingExplore(PingExplore):
...
 
+    def _to_lookml(self, v1_name: Optional[str]) -> dict:
+        """Generate LookML to represent this explore."""
+        repo = next((r for r in GleanPing.get_repos() if r["name"] == v1_name))
+        glean_app = GleanPing(repo)
+        ping_description = glean_app.get_ping_descriptions()[self.name]
+        return {
+            **super()._to_lookml(v1_name),
+            **{"description": f"Explore for the {self.name} ping. {ping_description}"},
+        }
```
fbertsch, line 26: I know this works but I think it may be extra confusing to look at. How do you feel about something like, just for clarity:
```suggestion
        lookml = super()._to_lookml(v1_name)
        lookml["description"] = f"Explore for the {self.name} ping. {ping_description}"
        return lookml
```

fbertsch (commented):
A few comments, let's make sure we deploy MSG and test this before merging.

generator/explores/glean_ping_explore.py
wlach: Yup. Python 3.9 will make this easier to do elegantly.

generator/lookml.py
wlach: Yup. I suspect the right way of dealing with v1_name is through changes to the probeinfo service and/or mozilla-schema-generator https://github.com/mozilla/lookml-generator/pull/143/commits/24386055db4638fa73cc8ea27a4d24455923f07c

tests/test_lookml.py
wlach: Things look pretty good in `looker-hub`: https://github.com/mozilla/looker-hub/blob/glean-ping-descriptions/fenix/explores/baseline.explore.lkml

I'm not sure what to look for in `looker-spoke-default`: https://github.com/mozilla/looker-spoke-default/tree/glean-ping-descriptions-working

generator/explores/glean_ping_explore.py
```diff
@@ -10,9 +15,36 @@ class GleanPingExplore(PingExplore):
...
+        """Generate LookML to represent this explore."""
+        repo = next((r for r in GleanPing.get_repos() if r["name"] == v1_name))
+        glean_app = GleanPing(repo)
+        # convert ping description indexes to snake case, as we already have
+        # for the explore name
+        ping_descriptions = {
+            k.replace("-", "_"): v for k, v in glean_app.get_ping_descriptions().items()
+        }
+        # collapse whitespace in the description so the lookml looks a little better
+        ping_description = " ".join(ping_descriptions[self.name].split())
```
wlach, line 28: I added this because the lookml looked bad with all the newlines and spaces. Should we do this in more places?

fbertsch: I think descriptions are the only place where this might happen. SQL statements we want to keep the whitespace as-is so it's more readable in the LookML files. We could add this logic to Glean Metric descriptions as well.

generator/explores/explore.py
```diff
@@ -15,6 +15,7 @@ class Explore:
     """A generic explore."""
 
     name: str
+    namespace: str
```
fbertsch, line 18: I'm not seeing where this field is being used. Can you point it out to me?

wlach: This actually turned out to be unnecessary, I removed it.

generator/explores/glean_ping_explore.py
wlach: I can look into doing this in a followup

fbertsch (approved):
Looks good! Let's try it out on stage

PR #151, opened 2021-06-03
Add autogenerated api docs using pdoc

acmiyaguchi:
This autogenerates api documentation, and embeds the README + architecture docs into the module documentation. I'm following Jeff's guide to deploying to circleci (https://circleci.com/blog/deploying-documentation-to-github-pages-with-continuous-integration/), and need to add the private key to circleci. The build issues with the google-api-core are blocking this though.

See: https://mozilla.github.io/lookml-generator/generator.html


fbertsch (approved):
This is a fine change, but changing the indentation of the yaml file was unexpected. I'd prefer we're consistent with indentation, can you update the [rules for indent-sequences](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.indentation)? 

I don't have a preference for either  or, but it should be `true` or `false` and we should maintain that for all the yaml files.

acmiyaguchi:
Sorry about that, I absentmindedly let prettier do it's thing and thought it would be okay. I'll enable the linting rule for consistency.

PR #158, opened 2021-06-08
Bump google-api-core to 1.30.0 (transitive dep)

acmiyaguchi:
See #152


fbertsch (approved):
https://en.wikipedia.org/wiki/Sisyphus

PR #161, opened 2021-06-11
Bug 1711119 - Add custom namespace for mozilla vpn

relud:


custom-namespaces.yaml
```diff
@@ -118,3 +118,39 @@ contextual_services:
       type: ping_explore
       views:
         base_view: topsites_impression
+mozilla_vpn:
```
fbertsch, line 121: Hm, we have a `-` `_` mismatch here. This will create a new `mozilla_vpn` directory. It looks like the model is called `mozilla_vpn.model.lkml`, maybe we should deploy this as-is and then copy everything in `mozilla-vpn` over to the new directory once it's deployed.

custom-namespaces.yaml
```diff
@@ -118,3 +118,39 @@ contextual_services:
...
+  glean_app: false
+  owners:
+    - dthorn@mozilla.com
+  pretty_name: Mozilla VPN
+  views:
```
fbertsch, line 126: Should we also add stripe tables?

relud: no, I have a different plan for those

custom-namespaces.yaml
relud: > maybe we should deploy this as-is and then copy everything in mozilla-vpn over to the new directory once it's deployed.

for consistency with other projects, i'm in favor of this approach

fbertsch (approved):

PR #162, opened 2021-06-14
Handle Glean pings with kebab case names (fixes #160)

### LINKED ISSUE #160: Some Glean pings are missing their metrics
The Credit Cards Sync ping has a few metrics. One is a counter [(outgoing batches)](https://dictionary.telemetry.mozilla.org/apps/fenix/metrics/creditcards_sync_outgoing_batches). However, following the link in the Glean Dictionary, the Looker dashboard is empty: https://mozilla.cloud.looker.com/explore/fenix/creditcards_sync?qid=YtzyGcDHQaMGfZ6WFyPq2O&toggle=vis

That explore doesn't show any metrics available at all, just some generic info from the Glean Ping itself. Clearly the metrics are not flowing through correctly. We would expect to see a few non-default Glean fields present in this explore.

wlach:
We need to be carefully about pings like `addresses-sync`, where the
explore name is snake case, but the original ping name is kebab.

### DIFF
--- a/tests/test_glean_ping_view.py
+++ b/tests/test_glean_ping_view.py
@@ -1,0 +1,70 @@
+from unittest.mock import Mock, patch
+
+from google.cloud import bigquery
+from google.cloud.bigquery.schema import SchemaField
+from mozilla_schema_generator.probes import GleanProbe
+
+from generator.views import GleanPingView
+
+
+class MockClient:
+    """Mock bigquery.Client."""
+
+    def get_table(self, table_ref):
+        """Mock bigquery.Client.get_table."""
+
+        if table_ref == "mozdata.glean_app.dash_name":
+            return bigquery.Table(
+                table_ref,
+                schema=[
+                    SchemaField(
+                        "metrics",
+                        "RECORD",
+                        fields=[
+                            SchemaField(
+                                "string",
+                                "RECORD",
+                                fields=[SchemaField("fun_string_metric", "STRING")],
+                            )
+                        ],
+                    ),
+                ],
+            )
+
+        raise ValueError(f"Table not found: {table_ref}")
+
+
+@patch("generator.views.glean_ping_view.GleanPing")
+def test_kebab_case(mock_glean_ping):
+    """
+    Tests that we handle metrics from kebab-case pings
+    """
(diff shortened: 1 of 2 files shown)


wlach:
This appears to fix things, tested a deploy on looker hub: https://github.com/mozilla/looker-hub/blob/wlach-ping-names/fenix/views/addresses_sync.view.lkml

Compare to: https://github.com/mozilla/looker-hub/blob/main/fenix/views/addresses_sync.view.lkml

@fbertsch I guess we probably want a unit test for this, but looking at tests/test_lookml.py kind of makes my head spin. Do you have any suggestions on how to proceed?

fbertsch:
@wlach I agree about a test case. I recommend creating a new file for just the Glean ping: `test_glean_ping_view.py`. In there you'll have to [patch MSG](https://github.com/mozilla/lookml-generator/blob/main/tests/test_lookml.py#L621). We should then be able to test with something like:
```
view = GleanPingView("glean_app", "dash_name", [{"channel": "release", "table": "mozdata.glean_app.dash_name"}]
lookml = view.to_lookml(mock_bq_client, "dash-name")
assert lookml contains "metric"
```

wlach:
Thanks Frank, that tip plus a bit of space from my computer gave me enough inspiration to write a viable test case. 😄

tests/test_glean_ping_view.py
```diff
@@ -0,0 +1,70 @@
...
+    ]
+    mock_glean_ping.return_value = glean_app
+    mock_bq_client = MockClient()
+    view = GleanPingView(
+        "glean_app",
+        "dash_name",
+        [{"channel": "release", "table": "mozdata.glean_app.dash_name"}],
+    )
+    lookml = view.to_lookml(mock_bq_client, "glean-app")
+    assert "metrics.string.fun_string_metric" in str(lookml)
```
fbertsch, line 70: Nit: When this check fails, it's often hard to understand why (if str(lookml) is a big blob). It is probably best to do an explicit check for the dimension.

fbertsch (approved):
r+wc. Thank you for writing the unit tests! That will be useful for more changes to the Glean Ping structure.

PR #168, opened 2021-06-22
Operational Monitoring Views + Explores

emtwo:
This really doesn't do what it needs to yet. I'm just putting up a WIP thats more of a template

### DIFF
--- a/generator/namespaces.py
+++ b/generator/namespaces.py
@@ -1,6 +1,7 @@
 """Generate namespaces.yaml."""
 import gzip
 import json
+import re
 import tarfile
 import urllib.request
 import warnings
@@ -14,12 +15,26 @@ from typing import Dict, List, Union
 
 import click
 import yaml
+from google.cloud import storage
 
 from .explores import EXPLORE_TYPES
 from .views import VIEW_TYPES, View
 
 PROBE_INFO_BASE_URI = "https://probeinfo.telemetry.mozilla.org"
 DEFAULT_SPOKE = "looker-spoke-default"
+OPMON_BUCKET_NAME = "operational_monitoring"
+PROD_PROJECT = "moz-fx-data-shared-prod"
+PROJECTS_FOLDER = "projects/"
+DATA_TYPES = {"histogram", "scalar"}
+
+
+def _normalize_slug(name):
+    return re.sub(r"[^a-zA-Z0-9_]", "_", name)
+
+
+def _download_json_file(project, bucket, filename):
+    blob = bucket.get_blob(filename)
+    return json.loads(blob.download_as_string()), blob.updated
 
 
 def _get_first(tuple_):
@@ -60,6 +75,52 @@ def _get_db_views(uri):
                         ref.split(".") for ref in references["view.sql"]
                     ]
     return views
+
+
+def _append_view_and_explore_for_data_type(
+    om_views_and_explores, project_name, table_prefix, data_type, branches
+):
+    project_data_type_id = f"{project_name}_{data_type}"
+
+    om_views_and_explores["views"][project_data_type_id] = {
+        "type": f"operational_monitoring_{data_type}_view",
+        "tables": [
+            {
+                "table": f"{PROD_PROJECT}.operational_monitoring.{table_prefix}_{data_type}"
+            }
+        ],
+    }
+    om_views_and_explores["explores"][project_data_type_id] = {
+        "type": "operational_monitoring_explore",
+        "views": {"base_view": f"{project_name}_{data_type}"},
+        "branches": branches,
+    }
+
+
+def _get_opmon_views_and_explores():
+    client = storage.Client(PROD_PROJECT)
+    bucket = client.get_bucket(OPMON_BUCKET_NAME)
+    om_views_and_explores = {"views": {}, "explores": {}}
+
+    # Iterating over all defined operational monitoring projects
+    for blob in bucket.list_blobs(prefix=PROJECTS_FOLDER):
+        # The folder itself is not a project file
+        if blob.name == PROJECTS_FOLDER:
+            continue
+
+        om_project, project_last_modified = _download_json_file(
+            PROD_PROJECT, bucket, blob.name
+        )
+        table_prefix = _normalize_slug(om_project["slug"])
+        project_name = om_project["name"].lower()
+        branches = om_project.get("branches", ["enabled", "disabled"])
+
+        for data_type in DATA_TYPES:
+            _append_view_and_explore_for_data_type(
+                om_views_and_explores, project_name, table_prefix, data_type, branches
+            )
+
+        return om_views_and_explores
 
 
 def _get_glean_apps(
@@ -207,6 +268,8 @@ def namespaces(custom_namespaces, generated_sql_uri, app_listings_uri, disallowl
 
     if custom_namespaces is not None:
         custom_namespaces = yaml.safe_load(custom_namespaces.read()) or {}
+        views_and_explores = _get_opmon_views_and_explores()
+        custom_namespaces["operational_monitoring"].update(views_and_explores)
         _merge_namespaces(namespaces, custom_namespaces)
 
     disallowed_namespaces = yaml.safe_load(disallowlist.read()) or {}
--- a/generator/views/operational_monitoring_view.py
+++ b/generator/views/operational_monitoring_view.py
@@ -1,0 +1,41 @@
+"""Class to describe an Operational Monitoring View."""
+
+from typing import Any, Dict, List, Optional
+
+from .ping_view import PingView
+from .view import ViewDict
+
+
+class OperationalMonitoringView(PingView):
+    """A view on a operational monitoring table."""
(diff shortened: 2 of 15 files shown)


generator/namespaces.py
```diff
@@ -14,12 +15,25 @@
...
+from google.cloud import storage
 
 from .explores import EXPLORE_TYPES
 from .views import VIEW_TYPES, View
 
 PROBE_INFO_BASE_URI = "https://probeinfo.telemetry.mozilla.org"
 DEFAULT_SPOKE = "looker-spoke-default"
+OPMON_BUCKET_NAME = "operational_monitoring"
+PROD_PROJECT = "moz-fx-data-shared-prod"
+PROJECTS_FOLDER = "projects/"
```
fbertsch, line 27: One tactic that we've really been leaning on is to expose metadata and configuration files in GIthub instead, where it can be world readable. It does ease debugging and would not require us to have GCP creds here. See if you think it would be worth making that change.

generator/views/operational_monitoring_view.py
```diff
@@ -0,0 +1,102 @@
...
+            if field.name in EXCLUDED_FIELDS:
+                continue
+
+            self.dimensions.append(
+                {
+                    "name": field.name,
+                    "type": field.field_type,
+                    "sql": "${TABLE}." + f"{field.name}",
+                }
+            )
```
fbertsch, line 81: This should either be using:
1. The utils in https://github.com/mozilla/lookml-generator/blob/main/generator/views/lookml_utils.py. See e.g. `ping_view` for how this works
2. Create a `table_view` for each one of these, and extend that view for these views. Then you only need to add the measures (and any derived dimensions). See e.g. `growth_accounting` for an example of this

generator/views/operational_monitoring_view.py
```diff
@@ -0,0 +1,102 @@
...
+        return {
+            "views": [
+                {
+                    "name": "fission    ",
+                    "derived_table": {
+                        "sql": dedent(
+                            f"""
+                            SELECT *
+                            FROM `{reference_table}`
+                            CROSS JOIN UNNEST(metrics)
```
fbertsch, line 92: This nested field is going to make generation difficult. Can we expose this in the view, so that generation can work on just the raw table?

fbertsch (commented):
A few thoughts so far. We should add a separate test file for this, like we did for [funnel analysis](https://github.com/mozilla/lookml-generator/blob/main/tests/test_funnel_analysis.py).

generator/namespaces.py
```diff
@@ -45,6 +59,30 @@ def _get_db_views(uri):
...
+            PROD_PROJECT, bucket, blob.name
+        )
+        table_name = _normalize_slug(om_project["slug"])
+        om_views[om_project["name"]] = {
+            "type": "operational_monitoring_view",
+            "tables": [
+                {"table": f"{PROD_PROJECT}.operational_monitoring.{table_name}"}
+            ],
+        }
+        return om_views
```
robhudson, line 83: All the `om`s make this code seem very meditative. :lotus_position:

generator/views/operational_monitoring_view.py
```diff
@@ -0,0 +1,102 @@
...
+    @classmethod
+    def from_dict(klass, namespace: str, name: str, _dict: ViewDict) -> PingView:
+        """Get a OperationalMonitoringView from a dict representation."""
+        return OperationalMonitoringView(namespace, name, _dict["tables"])
+
+    def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
+        """Get this view as LookML."""
+        if len(self.tables) == 0:
+            raise Exception(
+                (f"Operational Monitoring view {self.name} " f"has no tables")
```
robhudson, line 65: Probably don't need the 2 f-strings concatenated together here.

generator/views/operational_monitoring_view.py
```diff
@@ -0,0 +1,102 @@
...
+        schema = bq_client.get_table(reference_table).schema
+        for field in schema:
+            if field.name in EXCLUDED_FIELDS:
+                continue
+
+            self.dimensions.append(
+                {
+                    "name": field.name,
+                    "type": field.field_type,
+                    "sql": "${TABLE}." + f"{field.name}",
```
robhudson, line 79: Could be:
* `"${TABLE}." + field.name`, or
* `f"${{TABLE}}.{field.name}"` ?

generator/namespaces.py
emtwo: Well, right now this is subject to change anyways since we're only temporarily using data from a bucked until we decide what the user interface for creating new operational monitoring projects will look like. It's all set up this way on the bigquery side already now too. So I don't think it's worth the change at the moment but I will keep this in mind as we're designing the new intake process.

generator/views/operational_monitoring_view.py
emtwo: ok I used 1)

generator/views/operational_monitoring_view.py
emtwo: Do you mind elaborating on what you mean here? Do you mean to expose it in the BQ view or in the looker view? What is difficult about this?

emtwo:
> A few thoughts so far. We should add a separate test file for this, like we did for [funnel analysis](https://github.com/mozilla/lookml-generator/blob/main/tests/test_funnel_analysis.py).

@fbertsch thanks for the review! I've added a test for this and made a couple other updates.

emtwo:
Generated looker-hub branch: https://github.com/mozilla/looker-hub/commits/emtwo-generation-test-1

generator/views/operational_monitoring_view.py
```diff
@@ -0,0 +1,95 @@
...
+from typing import Any, Dict, List, Optional, Set
+
+from . import lookml_utils
+from .ping_view import PingView
+from .view import ViewDict
+
+# These are fields we don't need in our view
+EXCLUDED_FIELDS: Set[str] = {
+    "submission",
+    "client_id",
```
fbertsch, line 13: Any chance we can get `sample_id` in here? That is a nice one to make available for faster iterative analysis.

fbertsch (commented):
Next step is definitely to generate some explores. That would help us try and use these in Looker (right now it's just a static view file without any associated explores).

generator/views/operational_monitoring_view.py
emtwo: You mean to exclude `sample_id` or to include it? This is the exclude list. The table I'm prototyping with in BQ doesn't have a `sample_id` at the moment but this would require a change on the bq_etl side when a json config for operational monitoring is processed, then `sample_id` should be added in if it exists. I filed an issue for this change: https://github.com/mozilla/bigquery-etl/issues/2188

generator/views/operational_monitoring_view.py
emtwo: So I actually don't want to include these other dimensions in the looker view since they wouldn't be needed. Perhaps I'm conflating what would be in the looker view vs. what would be in the explore. The explore wouldn't expose these extra dimensions so I didn't see a need to have them in the view either.

generator/views/operational_monitoring_view.py
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    "name": "fission",
+                    "derived_table": {
+                        "sql": dedent(
+                            f"""
+                            SELECT *
+                            FROM `{reference_table}`
+                            CROSS JOIN UNNEST(metrics)
+                            WHERE PARSE_DATE('%Y%m%d', CAST(build_id AS STRING)) >
+                                DATE_SUB(CURRENT_DATE, interval 60 day)
```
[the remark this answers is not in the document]
emtwo: Do you mean the whole derived table should be done in the explore or you mean specifically for this `where` clause?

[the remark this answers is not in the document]
fbertsch: Basically we should let Looker insert the where clause. You can [require filters on explores](https://info.looker.com/data-modeling/look-learn-requiring-filters-on-explores), which still lets the user change it (to see e.g. just newer data, or data from more than 60 days ago).

generator/views/operational_monitoring_view.py
fbertsch: Yes sorry - I commented here because I saw `client_id`. We should include sample_id.

emtwo:
I've added explores and tests for the views + explores. Here is the branch with the looker hub output: https://github.com/mozilla/looker-hub/tree/emtwo/test2

The end goal is to mimic the hand-generated dashboard [here](https://mozilladev.cloud.looker.com/dashboards-next/fission::fission_with_dimensions?OS=Windows&Cores=4&Percentile=50) so it can be automated for any opmon project.

This PR starts by automating the generation of views and explores.

scholtzan (approved):
Nice! Looks good to me.
Once this is merged, we'll have to tag a new release, update the Airflow version variable, and then probe scraper will automatically create the views and explores.

PR #169, opened 2021-06-23
Add event counts view

fbertsch:
This currently won't create any new views for any datasets, since we haven't switched to the app-datasets for Glean apps.

### DIFF
--- a/generator/views/view.py
+++ b/generator/views/view.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 from typing import Any, Dict, Iterator, List, Optional, TypedDict
 
-import click
+from click import ClickException
 
 OMIT_VIEWS = {"deletion_request"}
 
--- a/generator/views/view.py
+++ b/generator/views/view.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 
 from typing import Any, Dict, Iterator, List, Optional, TypedDict
 
+import click
+
 OMIT_VIEWS = {"deletion_request"}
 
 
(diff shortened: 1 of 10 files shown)


generator/views/view.py
```diff
@@ -3,6 +3,8 @@
 
 from typing import Any, Dict, Iterator, List, Optional, TypedDict
 
+import click
```
wlach, line 6: Maybe just import the exception class that you need here, for clarity?

```suggestion
from click import ClickException
```

tests/utils.py
```diff
@@ -74,3 +78,16 @@ def print_and_test(expected, result=None, actual=None):
     print("\n".join([" - ".join(v) for v in get_differences(expected, result)]))
 
     assert result == expected
+
+
+def get_mock_bq_client(schema: List[SchemaField]):
```
wlach, line 83: 👍 Can you update tests/test_glean_ping_view.py to use this as well (fine if it's in a seperate PR)?

wlach (changes requested):
Looks good, however, netlify CI is failing. Can you take a look?

tests/utils.py
fbertsch: Yup, I'll add it to the backlog of refactoring 😄  My goal in general is to make it simpler to add new generated views and explores. We're pretty close to having it just be a few new files rather than changing a bunch of existing ones

fbertsch:
@wlach this is now a backwards-incompatible change. When we merge this we'll drop the `events` views for Glean apps. Before we deploy this to prod we'll have to ensure no one has built anything on those, and we'll replace them with the `events_unnested` views (which this event counts explore is built on).

wlach (approved):

PR #189, opened 2021-07-30
Update dev branches after looker-hub deploy

### LINKED ISSUE #65: Deploy `looker-hub` changes to all dev branches
We're running into a problem where changes to `looker-hub` will cause errors for developers in the Looker IDE. What's happening is that while they may see the update to `spoke-default` (by pulling them from production), they will not see the related changes to `looker-hub` unless they switch to that project and pull from production.

At this point there shouldn't be any reason for developers to make changes to `looker-hub` on their dev branch, since it's all automated. We can solve this issue by also pushing to every developer branch during deploy.

fbertsch:
Fixes #65.


bin/generate
```diff
@@ -186,6 +189,22 @@ function generate_hub_commit() {
...
+  cd /app/looker-hub
+
+  dev_branches_file="/app/lookml-generator/bin/dev_branches"
+  while read branch; do
+    git checkout $branch
+    git reset --hard main
+    git push -f origin $branch
+  done < $dev_branches_file
+
+  popd
```
jasonthomas, line 205: Are we using `pushd/popd` so that we can return the previous working directory?

jasonthomas (approved):

bin/generate
fbertsch: Yup, just a pattern I picked up from MSG.

PR #190, opened 2021-08-03
Add revenue namespace

akkomar:


fbertsch (approved):
LGTM. We've never had a namespace with no generated views/explores before, so let's keep an eye on it during deployment. Details here: https://github.com/mozilla/lookml-generator#deploying-new-lookml-generator-changes

Issue #94, opened 2021-05-11
Expose unnested `events` views rather than ping tables


fbertsch:
This depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1710088#c5.

We want to do two things here:
1. Stop outputting an `events` `GleanPing` explore and view. The events `TableView` is fine.
2. Create a new view and explore that extends `GleanPing`, but is unnested events, rather than metrics and nested events. Initially this will be named `events_unnested` in BQ.
3. Eventually, transition `events_unnested` to replace `events`, and also transition this view definition to instead read from the `events` table

fbertsch:
The above bug is complete. This issue is blocked on https://github.com/mozilla/lookml-generator/issues/156.

wlach:
This was fixed a while ago.

PR #200, opened 2021-08-12
Use app dataset on release channel for Glean (fixes #156)

### LINKED ISSUE #156: Glean applications should use app-dataset as default
All Glean applications now have an app dataset (distinct from their app-_id_ dataset). We want to expose this dataset instead of their default dataset (where "default dataset" is the release channel dataset for apps that have channels).

If this is easy, it should be as simple as replacing the `bq_dataset_family` for the release channel with the app name (during `namespaces.yaml` generation). That would then propagate to the lookml generation.

wlach:


wlach:
Hmm, I might be misunderstanding something but I think I implemented what the issue asked for and am getting some strange looking results when I look at the generated `namespaces.yaml`. @fbertsch do you have any ideas?

https://gist.github.com/wlach/3fab8322e351870ff22393b4a6f37430

wlach:
> Hmm, I might be misunderstanding something but I think I implemented what the issue asked for and am getting some strange looking results when I look at the generated `namespaces.yaml`. @fbertsch do you have any ideas?
> 
> https://gist.github.com/wlach/3fab8322e351870ff22393b4a6f37430

Frank actually explained why this was the case in slack (just missed it):

> When I made that issue, I thought the app-dataset had all the same views as the app-id-dataset, with the additional events_unnested view
> frank:sailboat:  17 hours ago
> however, that is not the case - it only has the events_unnested views. The other release views have not yet propagated to the app-dataset

wlach:
Zoomed with Frank. Next step for me is to update bigquery-etl to generate the necessary views here.

wlach:
> Zoomed with Frank. Next step for me is to update bigquery-etl to generate the necessary views here.

I think this might be more complicated than expected. It looks like for most of the stable views the views are being created based solely on the metadata created by mozilla-pipeline-schemas:

https://github.com/mozilla/bigquery-etl/blob/b92d4eb6711bdc7e130a95fc4c1cd94d2a4188b0/bigquery_etl/view/generate_stable_views.py#L169

Example of this metadata: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/0905f8fe66e546fb453d8f54519c761fee52b9ba/schemas/org-mozilla-firefox/activation/activation.1.schema.json):

If I'm not misunderstanding things, I see two alternatives:

1. Add some additional metadata in mozilla-pipeline-schemas to identify some alternate dataset family for these per-app datasets.
2. Make the above code in bigquery-etl read in the app listings yet again from probeinfo and merge this information in.

(1) seems like the better approach to me, though I'm not confident I understand things well enough here to say for sure.

scholtzan:
AFAIK what needs to happen here is similar to generating the per-app views we already have:
The existing per-app datasets get generated via `glean_usage`: https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/cli/glean_usage.py

The logic for that is in: https://github.com/mozilla/bigquery-etl/tree/main/bigquery_etl/glean_usage

So I wonder if it would make sense to move the per-app dataset generation into `glean_usage`.

We'll first need to know what stable tables there are in the dataset. We should be able to get them from the schemas. There is already [`list_baseline_tables()`](https://github.com/mozilla/bigquery-etl/blob/5098f9337f707670f2e75e4069f2005713cc14c3/bigquery_etl/cli/glean_usage.py#L83) which returns the baseline tables. I imagine we could add a method returning non-baseline tables.

Once we know what tables exist, we'll need to extend [`generate`](https://github.com/mozilla/bigquery-etl/blob/5098f9337f707670f2e75e4069f2005713cc14c3/bigquery_etl/cli/glean_usage.py#L113) and create the `UNION` views. In an ideal would it should be possible to simply re-use the [`cross_channel.view.sql`](https://github.com/mozilla/bigquery-etl/blob/main/bigquery_etl/glean_usage/templates/cross_channel.view.sql) template that already exists. However, schemas of tables across channels might differ, for example that's the case for `metrics`, so we'll probably need to determine the superset of all fields from the tables. See https://bugzilla.mozilla.org/show_bug.cgi?id=1712332

So I think we might want to add a new class (maybe `GleanPingTable`) for handling the generation and a new template in https://github.com/mozilla/bigquery-etl/tree/main/bigquery_etl/glean_usage. I don't think we need to add support for generating per-appId datasets here.

wlach:
Thanks @scholtzan. That all makes sense, though my understanding from #156 (and my conversation with @fbertsch about it) was that we were going to start just by adding an additional set of views for the *release channel only* indexed by app name, rather than app id. Unified per-app datasets would be a later thing. Let me double check with @fbertsch.

fbertsch:
> Thanks @scholtzan. That all makes sense, though my understanding from #156 (and my conversation with @fbertsch about it) was that we were going to start just by adding an additional set of views for the _release channel only_ indexed by app name, rather than app id. Unified per-app datasets would be a later thing. @fbertsch, can you confirm?

Exactly. It would make sense to first add just the release channel views. If you're willing and able to figure out the `UNION` view logic, that would be a nice next step but is not strictly necessary to complete this task.

wlach:
Ok, with mozilla/bigquery-etl#2271 landed, I think we're much closer:

https://gist.github.com/wlach/22ed672acb7948cacbd11f54252538be

It looks like a few things are still getting taken out: `funnel_analysis` and `migrated_clients`. I'm guessing these are just things we need to rename in bigquery-etl?

It also appears that in views we're still losing the release channel for the base ping view (e.g. `metrics`) although the table view (e.g. `metrics_table`) still needs to be picked up. I'm guessing this PR is probably just incomplete at this point.

fbertsch:
> Ok, with [mozilla/bigquery-etl#2271](https://github.com/mozilla/bigquery-etl/pull/2271) landed, I think we're much closer:
> 
> https://gist.github.com/wlach/22ed672acb7948cacbd11f54252538be
> 
> It looks like a few things are still getting taken out: `funnel_analysis` and `migrated_clients`. I'm guessing these are just things we need to rename in bigquery-etl?
> 
> It also appears that in views we're still losing the release channel for the base ping view (e.g. `metrics`) although the table view (e.g. `metrics_table`) still needs to be picked up. I'm guessing this PR is probably just incomplete at this point.

Good points - I didn't think about this aspect. We should probably replicate those views in the app-datasets in `bigquery-etl` (instead of creating more automation to replicate them).

wlach:
> Good points - I didn't think about this aspect. We should probably replicate those views in the app-datasets in `bigquery-etl` (instead of creating more automation to replicate them).

Can you clarify what you mean? I'm a bit confused about what needs to be done over and above what has happened already. See for example in this chunk from the diff I posted:

```
     metrics:
       tables:
-      - channel: release
-        table: mozdata.org_mozilla_firefox.metrics
       - channel: beta
         table: mozdata.org_mozilla_firefox_beta.metrics
```

These per-app tables (really views?) should be in there thanks to mozilla/bigquery-etl#2271 -- although the deployment issues with https://github.com/mozilla/bigquery-etl/issues/2214 (which I wasn't aware of when I wrote the above) appear to be blocking them from being deployed.

wlach:
Update: I looked into this more deeply and the tables aren't being picked up because lookml-generator doesn't think they reference stable ping tables:

https://github.com/mozilla/lookml-generator/blob/3143001e0692ac60218e8c9a6d20a6a0681d9371/generator/views/ping_view.py#L45

Specifically note the references for the new ping table view:

```
# Generated by bigquery_etl.dependency
references:
  view.sql:
  - moz-fx-data-shared-prod.org_mozilla_firefox.activation
```
https://github.com/mozilla/bigquery-etl/blob/83bd924949d4c4bfbd381ae602979affbb6e00fd/sql/moz-fx-data-shared-prod/fenix/activation/metadata.yaml#L15

I am tempted to suggest that we should augment bigquery-etl with some logic to mark specific views as referencing ping tables (e.g. with a `ping_table: true` property) and then pulling that information out in lookml-generator around here:

https://github.com/mozilla/lookml-generator/blob/3143001e0692ac60218e8c9a6d20a6a0681d9371/generator/namespaces.py#L43

Don't know if that's reasonable or if there's a better way. @scholtzan @fbertsch any thoughts?

wlach:
P.S. A `ping_table` property might also be useful for generating better documentation inside bigquery-etl.

scholtzan:
We _might_ just get away with resolving references without views in https://github.com/mozilla/bigquery-etl/blob/b5b10ec5bb95d79e4fde841b12037f5cc0f3f4fe/bigquery_etl/dependency.py#L187

So it would be `_get_references(paths, True)`. I've looked through `references` usages and I don't think it should break anything, but I might be missing something here.

wlach:
> We _might_ just get away with resolving references without views in https://github.com/mozilla/bigquery-etl/blob/b5b10ec5bb95d79e4fde841b12037f5cc0f3f4fe/bigquery_etl/dependency.py#L187
> 
> So it would be `_get_references(paths, True)`. I've looked through `references` usages and I don't think it should break anything, but I might be missing something here.

Nevermind, I found an easier solution: these new ping tables only have one reference (which is already a condition we're looking for), so we can just expand our search to look for `dataset` as well as `dataset_$stable`.

https://github.com/mozilla/lookml-generator/pull/200/commits/fd394633a630afc36046abe95aa60e6789bda2c4#diff-15a4bb2f6ffeceecd4adb1ae4b09afe0308d51de25ce5e869087f80ec1b8f06bR47

For robustness, I do think we're eventually going to want to generate a better definition of a "ping table" in bigquery etl but I think this gets us what we need for now.

Unfortunately there are still a number of odd-tables-out like `funnel_analysis`. I think the solution there is to dig back into bigquery-etl.

Here's the latest diff of namespaces.yaml (we should probably generate this automatically):

https://gist.github.com/wlach/eac3fcc62d8fc9ccb7a21f73e397b4cb

generator/views/ping_view.py
```diff
@@ -42,7 +42,10 @@ def from_db_views(
 
                 if channel.get("channel") is not None:
                     table["channel"] = channel["channel"]
-                if len(references) != 1 or references[0][-2] != f"{dataset}_stable":
+                if len(references) != 1 or references[0][-2] not in [
+                    f"{dataset}_stable",
+                    dataset,
```
scholtzan, line 47: Wouldn't the change in `namespaces.py` set result in this being the app dataset, e.g. `fenix` (which won't be referenced by the view)?

Thinking ahead a little more, at some point when we'll have the `UNION` app dataset views, this condition won't hold anymore since we'll be referencing multiple tables. So adding something like `ping_table: true` to the metadata is probably the most convenient path forward

scholtzan:
> Can you clarify what you mean? I'm a bit confused about what needs to be done over and above what has happened already. 

I think the easiest solution and what Frank means is to add the `view.sql` definitions for these views in bigquery-etl manually. I looks like there are only a couple. So we'd have, for example, `fenix/migrated_clients/view.sql`

generator/views/ping_view.py
wlach: > Wouldn't the change in `namespaces.py` set result in this being the app dataset, e.g. `fenix` (which won't be referenced by the view)?

No, because the `dataset` variable here is referring to the dataset we extracted out of the references, which *is* the app-id-specific dataset (e.g. `org_mozilla_firefox`).

wlach:
> > Can you clarify what you mean? I'm a bit confused about what needs to be done over and above what has happened already.
> 
> I think the easiest solution and what Frank means is to add the `view.sql` definitions for these views in bigquery-etl manually. I looks like there are only a couple. So we'd have, for example, `fenix/migrated_clients/view.sql`

That's kinda what I figured. I'm going to try to get that working, then we can circle around and figure out what the best strategy is for getting this PR landed (i.e. whether to land as-is or add more things to bigquery-etl).

wlach:
I'm pretty sure https://github.com/mozilla/bigquery-etl/pull/2286 takes care of most of this. There's still some missing ping views for release in my latest diff, but I *think* that is due to the deployment problems we've been having with the views themselves (see https://github.com/mozilla/bigquery-etl/issues/2214). I'll try to confirm that while waiting for other things to land.

generator/views/ping_view.py
wlach: I'm not sure what I was thinking when I wrote the above, but I was wrong and the above code doesn't work. I added a new concept called a `source_dataset` (the app-id-level view for the case of the release channel, the stable table in the case of non-release) which we can use here to work around this for now.

tests/conftest.py
```diff
@@ -36,7 +36,7 @@ def app_listings_uri(tmp_path):
                         "app_name": "glean-app",
                         "app_channel": "release",
                         "canonical_app_name": "Glean App",
-                        "bq_dataset_family": "glean_app",
+                        "bq_dataset_family": "glean_app_release",
```
wlach, line 39: This more realistically maps to what we have in production, at least for Fenix.

wlach:
I think this is ready for a serious review now. This won't actually work until the above mentioned deploy issues are fixed (and might need one or two other revisions afterwards) but now would probably be the time to say "stop! don't do it like this" 😄

scholtzan:
This looks good 👍 https://github.com/mozilla/bigquery-etl/issues/2287 is also related to deployment failures and missing views.

wlach:
With various issues in bigquery-etl fixed, the new namespaces.yaml is looking great. Here's a diff of what we have vs. production: https://gist.github.com/wlach/dd9233b606ec4890ae6a89d3becc0cf7

Now we just need to verify that the new events explores are actually accessible and useful. 🤞

scholtzan (approved):
This looks good to me. I've pushed the changes to https://github.com/mozilla/looker-hub/tree/app-dataset-generation-test and checked them in Looker.

I think this is good to get merged and will unblock https://github.com/mozilla/lookml-generator/issues/94 (the events unnested views support)

PR #201, opened 2021-08-13
Make sure primary key for Glean labeled counters is unique (fixes #171)

### LINKED ISSUE #171: Labeled counters have a non-unique primary key
The [primary key](https://docs.looker.com/reference/field-params/primary_key) for Glean Ping Labeled Counters is the [document id](https://github.com/mozilla/lookml-generator/blob/main/generator/views/glean_ping_view.py#L78). The problem is that this is not actually a primary key for the unnested join - each row in the labeled counter shares the same document id, [causing overflow](https://mozilla.cloud.looker.com/explore/firefox_ios/metrics?qid=wZIDSVE4oYJDhyuy9MWCws&toggle=fil,vis) when querying multiple counters.

Instead, we should create a new field that concatenates the doc-id and label, and set that as the primary key (and also hide it). That field will indeed be unique across all rows after the unnested join.

wlach:


generator/views/glean_ping_view.py
```diff
@@ -87,7 +87,7 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
                         {
                             "name": "document_id",
                             "type": "string",
-                            "sql": f"${{{self.name}.document_id}}",
+                            "sql": f"${{{self.name}.document_id}}-${{label}}",
```
scholtzan, line 90: I think we might want to keep `document_id` around as is and create a new dimension with this unique value. Otherwise this might cause confusion if the `document_id` has a different format here.

scholtzan (approved):
lgtm

wlach:
@fbertsch gave a 👍 on slack

PR #202, opened 2021-08-13
Pull in descriptions from BQ for dimensions (fixes #138)

### LINKED ISSUE #138: Pull in descriptions from BQ for all non-Glean dimensions
Most fields in BQ have an associated description. We should pull these out of BQ and insert them in to the Looker explores. That would happen in `/generator/views/lookml_utils.py`, during the `_get_dimension` function.  We would need to pull out the description in the parent function (`_generate_dimensions_helper`), and send it in as another parameter to `_get_dimension`. It would then add that as another field to the dimension:
```
result["description"] = "..."
```

scholtzan:
Fixes https://github.com/mozilla/lookml-generator/issues/138

Adds descriptions to dimensions if they are available in BQ

### DIFF
--- a/generator/views/lookml_utils.py
+++ b/generator/views/lookml_utils.py
@@ -1,6 +1,6 @@
 """Utils for generating lookml."""
 import re
-from typing import Any, Dict, Iterable, List, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 import click
 from google.cloud import bigquery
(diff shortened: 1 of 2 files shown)


fbertsch (commented):
Looks fine to me but I'd like Anthony to approve. One question is to make sure Glean ping dimensions and measures are correctly pulling in the descriptions.

acmiyaguchi (approved):
LGTM. Regard the glean ping views, it looks like they should already handle descriptions appropriately from the `_get_dimension` helper.

PR #208, opened 2021-08-17
Remove type from Glean Metric Dimensions (fixes #99)

### LINKED ISSUE #99: Glean metric dimensions should not have type in their name
For Glean metric dimensions, we use both `label` and `name`. The `name` is generally hidden from end users, but it comes up in the table that is displayed after a query is run. We should change the name to roughly match the metric name. [Code here](https://github.com/mozilla/lookml-generator/blob/main/generator/views/glean_ping_view.py#L96).

wlach:
This additional information isn't necessary, and looks redundant. For example: 

image

### DIFF
--- a/generator/views/glean_ping_view.py
+++ b/generator/views/glean_ping_view.py
@@ -249,17 +249,18 @@ class GleanPingView(PingView):
         if not group_label:
             group_label = "Glean"
 
+        friendly_name = f"{group_label} {group_item_label}"
+
         lookml = {
             "name": looker_name,
+            "label": friendly_name,
             "sql": sql_map[looker_name]["sql"],
             "type": sql_map[looker_name]["type"],
             "group_label": group_label,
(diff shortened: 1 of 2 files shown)


scholtzan (approved):

generator/views/glean_ping_view.py
```diff
@@ -216,16 +216,12 @@ def _get_category_and_name(self, metric: GleanProbe) -> Tuple[str, str]:
 
         return category, name
 
-    def _to_looker_name(self, metric: GleanProbe, suffix: str = "") -> str:
+    def _to_looker_name(self, metric: GleanProbe) -> str:
```
fbertsch, line 219: This function is not used by the dimension names, only by the view names of the joined labeled counters. Options are to either use this in `_make_dimension`, or update that code as well.

scholtzan (approved):
🎉

PR #210, opened 2021-08-17
Check if Looker SDK is set up before initialization

scholtzan:
I tested it locally and the exception did not get raised.


wlach (approved):
💯

PR #216, opened 2021-08-23
Use namespaces-disallowlist.yaml (fixes #136)

### LINKED ISSUE #136: Move to a `namespaces-disallowlist.yaml`
Moving forward, we want to enable namespaces for all new Glean applications. To enable this, we should remove the allowlist and create a disallowlist. It will initially contain everything _not_ in the allowlist, which will allow new Glean apps to flow in to Looker. We can then start removing apps from the disallowlist.

scholtzan:
Fixes https://github.com/mozilla/lookml-generator/issues/136

Having a disallow list will enable namespaces for all new Glean applications. The first use case for this is FOG which can result in new apps being added (e.g. like `firefox_desktop_background_update`)

The [generated `namespace.yaml` file](https://deploy-preview-216--lookml-generator-dev.netlify.app/) is identical to the one currently in [looker-hub](https://github.com/mozilla/looker-hub/blob/main/namespaces.yaml)


generator/namespaces.py
```diff
@@ -26,6 +26,23 @@ def _get_first(tuple_):
     return tuple_[0]
 
 
+def _merge_namespaces(dct, merge_dct):
+    """Recursively merge namespaces."""
+    for k, _ in merge_dct.items():
+        if k in dct and isinstance(dct[k], dict) and isinstance(merge_dct[k], Mapping):
+            if "glean_app" in merge_dct[k] and merge_dct[k]["glean_app"] is False:
+                # if glean_app gets set to False, Glean views and explores should not be generated
+                dct[k] = merge_dct[k]
```
scholtzan, line 35: This is necessary when generating the `mozilla_vpn` namespace

relud (approved):

PR #233, opened 2021-09-03
Hide dimensions for Glean metrics that are no longer available (fixes #232)

### LINKED ISSUE #232: Hide dimensions for Glean metrics that are no longer available
Currently we should dimensions for glean pings that no longer exist, for example none of the credit card metrics here are in the source (they have been renamed to something else):

image

We'll be marking these metrics as such in the Glean Dictionary shortly (https://github.com/mozilla/glean-dictionary/issues/705) but we should also hide these dimensions by default in the Looker explore view. They'll still be accessible via the Glean Dictionary (and the URL) if people really need them. 

Marking a field as hidden in looker is pretty straightforward, I have a PR that I'll post shortly:

https://docs.looker.com/reference/field-params/hidden

wlach:


wlach:
This is failing because we need a new version of mozilla-schema-generator: https://github.com/mozilla/mozilla-schema-generator/pull/209

wlach:
Diff of a test run against current main looker-hub for fenix metrics, seems to line up with expectations (stuff that is no longer in source is now marked as hidden): https://gist.github.com/wlach/294cf99a508873b0a841661c914a5d01

(I manually corrected an error in an earlier PR where we were storing the hidden status as a tuple)

scholtzan (approved):
🎉

PR #237, opened 2021-09-20
Remove workaround for joshtemple/lkml#53

relud:


relud:
CI failure fixed in #238

tests/test_dump_query.py
```diff
@@ -0,0 +1,22 @@
+import lkml
+
+
+def test_dump_query():
+    # adapted from https://github.com/joshtemple/lkml/pull/55/files
```
wlach, line 5: Do we actually need this? Hopefully we can trust lkml not to break.

wlach (approved):

tests/test_dump_query.py
relud: i can drop this. I wanted to test that the fix is present as we remove the workaround.

PR #239, opened 2021-09-20
[Look-81] Add New Tab namespace

scholtzan:
https://mozilla-hub.atlassian.net/browse/LOOK-109

Adds the New Tab namespace

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -222,3 +222,10 @@ firefox_ios:
 fenix:
   owners:
     - vzare@mozilla.com
+newtab:
+  glean_app: false
+  connection: bigquery-oauth
+  pretty_name: New Tab
+  owners:
+    - mmccorquodale@mozilla.com
+  spoke: looker-spoke-private


custom-namespaces.yaml
```diff
@@ -222,3 +222,10 @@ firefox_ios:
 fenix:
   owners:
     - vzare@mozilla.com
+newtab:
+  glean_app: false
+  connection: bigquery-oauth
+  pretty_name: New Tab
+  owners:
+    - mmccorquodale@mozilla.com
+  spoke: looker-spoke-private
```
scholtzan, line 231: @mmccorks Do you already know what tables the dashboards will be based on? If so, we could add them here and Looker will automatically import them.

scholtzan: Ah nvm. I just took a look at the dev dashboard and it looks like all the views and explores it uses already got imported to other namespaces.

mmccorks (approved):
LGTM

PR #240, opened 2021-09-21
Firefox desktop namespace

scholtzan:
Generated views and explores: https://github.com/mozilla/looker-hub/tree/ff-desktop-test

### DIFF
--- a/bin/generate
+++ b/bin/generate
@@ -142,11 +142,11 @@ function check_files_and_commit() {
   # Check that base branch files are unchanged. Error if they are.
   # Add the new files and commit.
 
-  if ! git diff-index --quiet HEAD;
-  then
-    git diff-index HEAD | cut -f 2 | xargs -I % echo "Error: lookml-generator modified %"
-    exit 1
-  fi
+  # if ! git diff-index --quiet HEAD;
+  # then
+  #   git diff-index HEAD | cut -f 2 | xargs -I % echo "Error: lookml-generator modified %"
+  #   exit 1
+  # fi
 
   # Use interactive mode to add untracked files
   # This also works when it's untracked directories
--- a/bin/generate
+++ b/bin/generate
@@ -142,11 +142,11 @@ function check_files_and_commit() {
   # Check that base branch files are unchanged. Error if they are.
   # Add the new files and commit.
 
-  # if ! git diff-index --quiet HEAD;
-  # then
-  #   git diff-index HEAD | cut -f 2 | xargs -I % echo "Error: lookml-generator modified %"
-  #   exit 1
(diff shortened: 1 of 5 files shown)


bin/generate
```diff
@@ -142,11 +142,11 @@ function check_files_and_commit() {
...
-  if ! git diff-index --quiet HEAD;
-  then
-    git diff-index HEAD | cut -f 2 | xargs -I % echo "Error: lookml-generator modified %"
-    exit 1
-  fi
+  # if ! git diff-index --quiet HEAD;
+  # then
+  #   git diff-index HEAD | cut -f 2 | xargs -I % echo "Error: lookml-generator modified %"
+  #   exit 1
+  # fi
```
scholtzan, line 149: There are a couple of [views in lookerhub for firefox_desktop](https://github.com/mozilla/looker-hub/tree/main/firefox_desktop/views) that have been created manually. Before merging this PR, the views need to be deleted. We'll only use the auto-generated desktop views in the future.

wlach (changes requested):
This is exciting! I think my main piece of feedback right now is that we should somehow make it more clear that the `metrics` explore for Firefox Desktop and all of the background updater stuff are "Glean". This gives a (small) hint that they are different from the other explores, which use the battle-tested legacy telemetry.

I think we can do this just by adding a Glean in parans next to the name. i.e. `Metrics (Glean)` for Firefox desktop and `Firefox Desktop Background Updater (Glean)` for the background updater. Maybe @chutten has additional thoughts?

I wouldn't mind doing a bit of light testing on this on dev/stage before it lands too.

scholtzan:
Agreed, that makes sense. I added `Telemetry` and `Glean` labels to the explores in https://github.com/mozilla/looker-spoke-default/pull/145/files#diff-fc367b99561996f41002784988bc3c33f4121ab35bd76294d29f2a13374df368R14 I don't think we need to add support to lookml-generator for automatically adding these labels temporarily.

wlach (approved):

wlach:
> If none of the other Glean-based explores from non-desktop projects are marked `(Glean)`, maybe we should instead mark the Telemetry-based views as `(Telemetry)`. They are the ones that are temporary (for a multi-year definition of "temporary") after all.

@scholtzan marked both in the spoke, which I think is probably the right approach for now. At some point, we can stop treating Glean as a special case but for now I think it's best to clearly mark explores derived from this data.

scholtzan:
Yes, existing views and explores can be updated using the `+<name>` syntax. Like:

```
explore: +growth_accounting {...}
view: +growth_accounting_table {...}
```

PR #241, opened 2021-09-21
Add Awesome Bar namespace

scholtzan:
https://mozilla-hub.atlassian.net/browse/LOOK-113


custom-namespaces.yaml
```diff
@@ -229,3 +229,13 @@ newtab:
...
+awesome_bar:
+  glean_app: false
+  owners:
+    - tbrooks@mozilla.com
+  pretty_name: Awesome Bar
+  views:
+    urlbar_clients_daily:
+      type: ping_view
+      tables:
+        - table: mozdata.telemetry.urlbar_clients_daily
```
scholtzan, line 241: @teonbrooks Are there any other tables you'd need?

teonbrooks: @scholtzan nope, that's it :)

teonbrooks: just curious, i see that the type is `ping_view`, what does that mean?

custom-namespaces.yaml
```diff
@@ -229,3 +229,13 @@ newtab:
...
     - mmccorquodale@mozilla.com
   spoke: looker-spoke-private
+awesome_bar:
+  glean_app: false
+  owners:
+    - tbrooks@mozilla.com
+  pretty_name: Awesome Bar
+  views:
+    urlbar_clients_daily:
+      type: ping_view
```
teonbrooks, line 239: just curious, i see that the type is `ping_view`, what does that mean?

akkomar: I'm not familiar with this, but looking at https://github.com/mozilla/lookml-generator/tree/main/generator/views this seems to be the right parser for this table. Though this wasn't obvious to me at first just reading its name.

akkomar (approved):

custom-namespaces.yaml
scholtzan: This could have also been a `table_view`, but `ping_view` works here too since it has a `client_id` and some `channel`. Using a `ping_view` some additional measures, like a count of clients, are auto generated

PR #242, opened 2021-09-21
Add UPDATE_SPOKE_BRANCHES option to generate

scholtzan:
Setting `$UPDATE_SPOKE_BRANCHES=false` will prevent lookml-generator from opening PRs against the spoke projects. This comes in handy when doing local testing.


README.md
```diff
@@ -84,9 +84,14 @@ a new branch to the `looker-hub` repository:
 ```
 export HUB_BRANCH_PUBLISH="yourname-generation-test-1"
 export GIT_SSH_KEY_BASE64=$(cat ~/.ssh/id_rsa | base64)
+export UPDATE_SPOKE_BRANCHES=false
+
 make build && make run
 ```
 
+To prevent PRs from automatically getting opened against spoke-default and spoke-private,
+while testing local changes, `UPDATE_SPOKE_BRANCHES` can be set to `false`.
```
wlach, line 93: I think this should be the default behaviour, which would remove the need for us to document it here as part of the developer workflow (and risk people making mistakes). This a little more effort (since it will require modifying our deployment logic) but I think it will be worth it in the long term.

wlach (changes requested):
Code looks right to me, but I think this should be the default behaviour. Let me know if there's something I'm missing.

README.md
scholtzan: Yeah, I think that makes sense.

scholtzan: I updated the Airflow task and set `UPDATE_SPOKE_BRANCHES` to false by default.

bin/generate
```diff
@@ -40,6 +40,8 @@
 #                             access to repos. Not required in dev.
 #   UPDATE_DEV_BRANCHES:      Whether or not the dev branches should be updated.
 #                             This should only happen in production.
+#   UPDATE_SPOKE_BRANCHES:    Whether or not pull-requests should be opened against
+#                             branches in spoke projects.
```
wlach, line 44:
```suggestion
#                             branches in spoke projects.
#                             This should only happen in production.
```

docker-compose.yml
```diff
@@ -22,3 +22,4 @@ services:
       - LOOKER_API_CLIENT_SECRET
       - GITHUB_ACCESS_TOKEN
       - UPDATE_DEV_BRANCHES
+      - UPDATE_SPOKE_BRANCHES
```
wlach, line 25: Does this need to be set?

wlach (approved):
🎉

docker-compose.yml
scholtzan: Yes, otherwise the [`UPDATE_SPOKE_BRANCHES` env variable](https://github.com/mozilla/telemetry-airflow/pull/1362/files#diff-67fcd8dc7d9c494d949102b2e5d1dc33e864059908bb5605144779f11101078fR156) that we set in the Airflow DAG would not get passed into the Docker container

PR #248, opened 2021-09-28
Remove views referencing experiment project

scholtzan:
Removed due to Airflow not having permission to access them. Will just add them manually to spoke


whd:
You could alternatively add airflow access via https://github.com/mozilla-services/data-sandbox-terraform/blob/main/projects/experiments-analysis/project/main.tf#L62

Issue #31, opened 2021-04-16
Handle nested fields


fbertsch:
Nested fields should be handled as joined tables, with dimensions and measures. We should also automate sub-explores that allow us to include suggestions for dimensions inside those generated joins.

PR #254, opened 2021-10-04
Handle nested fields

scholtzan:
Fixes https://github.com/mozilla/lookml-generator/issues/31


generator/explores/explore.py
```diff
@@ -105,6 +105,23 @@ def _get_default_channel(self, view: str) -> Optional[str]:
...
+        self, view_name: str, views: List[str]
+    ) -> Tuple[str, str]:
+        """
+        Get base view and metric names.
+
+        Returns the the name of the base view and the metric based on the
+        passed `view_name` and existing views.
+        """
+        split = view_name.split("__")
+        for index in range(len(split) - 1, 0, -1):
```
wlach, line 118: I assume you're working backward until you find a matching view, and we need to do it this way because there might be more than one view that matches? Can you add a quick comment describing the logic here?

generator/views/glean_ping_view.py
```diff
@@ -54,6 +54,8 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
         against the view in the explore.
         """
         lookml = super().to_lookml(bq_client, v1_name)
+        # ignore nexted join views
```
wlach, line 57: Typo: nexted -> nested. It would also probably be a good idea to expand on this comment (e.g. with the additional context below)

generator/views/lookml_utils.py
```diff
@@ -121,6 +121,44 @@ def _generate_dimensions(client: bigquery.Client, table: str) -> List[Dict[str,
     return list(dimensions.values())
 
 
+def _generate_nested_dimension_views(
+    schema: List[bigquery.SchemaField], view_name: str
+) -> List[Dict[str, Any]]:
+    """
+    Recursively generate views for nested fields.
+
+    Nested fields are handled as view, with dimensions and optionally measures.
```
wlach, line 130:
```suggestion
    Nested fields are created as views, with dimensions and optionally measures.
```

generator/explores/ping_explore.py
```diff
@@ -15,13 +15,33 @@ class PingExplore(Explore):
 
     def _to_lookml(self, v1_name: Optional[str]) -> List[Dict[str, Any]]:
         """Generate LookML to represent this explore."""
+        views_lookml = self.get_view_lookml(self.views["base_view"])
+        views = [view["name"] for view in views_lookml["views"]]
+
+        joins = []
```
wlach, line 21: Guess we may as well add a type annotation here?

```suggestion
        joins: List[Dict[str, str]] = []
```

tests/test_lookml.py
```diff
@@ -1754,7 +1780,26 @@ def test_context_id(runner, glean_apps, tmp_path):
...
+                    ],
+                    "name": "context__contexts",
+                },
+                {
+                    "dimensions": [
+                        {"name": "key", "sql": "${TABLE}.key", "type": "string"},
+                        {"name": "value", "sql": "${TABLE}.value", "type": "number"},
+                    ],
+                    "name": "context__contexts__position",
+                },
```
wlach, line 1802: Hmm, unless I'm missing something it looks like we're not actually testing explore generation in these tests? Might be good to add that (could be done as a followup).

wlach (approved):
Some concerns about testing, aside from that LGTM!

tests/test_lookml.py
scholtzan: Right, those are just the views. I'll add a test here for checking the generated joins in the explore

generator/explores/explore.py
scholtzan: Added an example to the function doc comment

PR #260, opened 2021-10-08
Allow user to select event extras (properties) in Funnel Analysis

wlach:
Fixes https://mozilla-hub.atlassian.net/browse/DS-1776


wlach:
Basically everything was already here, it just needed a bit of lookml to be wired up.

PR #264, opened 2021-10-11
Update custom-namespaces.yaml

xluo-ds:
Add a view for desktop search metric alert.


scholtzan:
You can just update it here :)

PR #269, opened 2021-10-13
Make Mozilla VPN a Glean app

wlach:
Primarily to enable the funnel analysis explore.


wlach:
@relud I'm still a bit of a n00b at this but attempted to generate a test branch. The lookml *looks* basically fine (aside from creating a bunch of duplicate `_table` views, but it's a bit hard to parse (it seems to think there's changes against main where there aren't any): https://github.com/mozilla/looker-hub/compare/wlach-vpn-generation-test-1

I assume this is just technical debt we can clean up later? There's a bit of a rush to be able to explore this data in Looker from what I understand.

relud:
yeah, looks like it shouldn't cause any issues

PR #273, opened 2021-10-19
Add monitoring namespace

scholtzan:
Adding a new `monitoring` namespace. I also added some of the views that we will probably need for migrating the existing dashboards over to Looker at some point.


custom-namespaces.yaml
```diff
@@ -421,3 +421,92 @@ sync:
       type: ping_explore
       views:
         base_view: sync
+monitoring:
+  glean_app: false
+  owners:
+    - jklukas@mozilla.com
+    - ascholtz@mozilla.com
+    - wlach@mozilla.com
+  pretty_name: Monitoring
```
jklukas, line 430:
```suggestion
  pretty_name: Monitoring (Data Pipeline)
```

Given the visibility of "Operational Monitoring" as a concept, it seems like we should provide a bit more context about what this is.

PR #295, opened 2021-11-09
Add aggregate tables to speed up operational monitoring dashboards.

emtwo:
This PR adds aggregate tables to Operational Monitoring explores to speed up their loading. [This branch](https://github.com/mozilla/looker-hub/tree/emtwo/aggregate_tables) in looker-hub has the changes that can be tested.

In order to add an aggregate table for each metric, the set of metrics associated with an explore need to be available to it including that metric's default values so it can be placed in the aggregate table. This data was already being processed in the dashboard. 

In order to avoid doing the same processing twice - in explores and in dashboards, I factored out that processing in the dashboard and turned it into generic data functions that can be applied to any namespace, called only once to perform processing, and accessed in both explores and dashboards.

### DIFF
--- a/generator/explores/operational_monitoring_explore.py
+++ b/generator/explores/operational_monitoring_explore.py
@@ -1,10 +1,13 @@
 """Operational Monitoring Explore type."""
 from __future__ import annotations
 
+from copy import deepcopy
 from pathlib import Path
 from typing import Any, Dict, Iterator, List, Optional
 
-from ..views import View
+from google.cloud import bigquery
+
+from ..views import View, lookml_utils
 from . import Explore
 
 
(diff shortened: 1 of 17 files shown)


generator/explores/operational_monitoring_explore.py
```diff
@@ -45,16 +48,57 @@ def from_dict(
...
+    def _to_lookml(
+        self,
+        bq_client: bigquery.Client,
+        v1_name: Optional[str],
+        data: Dict = {},
+    ) -> List[Dict[str, Any]]:
+        base_view_name = self.views["base_view"]
+
+        namespace_data = data.get("compute_opmon_dimensions", {}).get(
+            base_view_name, {}
```
scholtzan, line 60: Instead of having to define `data_functions` and passing around `namespace_data`, I'm wondering if it is possible to move `compute_opmon_dimensions()` into the `OperationalMonitoringExplore` implementation. The explore does have access to all the `view` names and could pass them into `compute_opmon_dimensions()`.
The `compute_opmon_dimensions()` is pretty specific to operational monitoring, so it's unlikely we will want to reuse it elsewhere. Moving it out of `lookml_utils` (which mostly contains generic lookml functions) might be nice.

emtwo: The `namespace_data` gets passed into both the explore and the dashboard so I'm not sure `OperationalMonitoringExplore` is the right place for `compute_opmon_dimensions()`.

I think there could be a separate `utils` file for opmon so that `compute_opmon_dimensions()` doesn't need to stay in `lookmk_utils`. Since the data produced by that function is used by both explores and dashboards, I think it doesn't necessarily belong in either.

scholtzan: I see. I think this makes sense. I'm not super in love with having a new `data_functions` attribute, but it's okay for now. 

As we are moving towards generating `aggregate_tables` for other generated explores, we might want to refactor this and have a better way of making dimensions available across views, explores and dashboards (+ some way to filter/transform them). I'll write up an issue to keep track of this.

For now moving `compute_opmon_dimensions()` into a separate file would be great.

scholtzan (approved):

emtwo:
> As we are moving towards generating aggregate_tables for other generated explores, we might want to refactor this and
> have a better way of making dimensions available across views, explores and dashboards (+ some way to filter/transform
> them). I'll write up an issue to keep track of this.

I agree that I don't love the `data_functions` to be honest. I think it will be easier to design how this works once we do have another use-case of `aggregate_tables`

PR #305, opened 2021-11-19
Add a default y axis formatter for the funnel analysis explores

wlach:
This will show the percentages as actual percentages, instead of numbers
like 0.15%.


wlach:
Before:

image

After:

image

from: https://mozilla.cloud.looker.com/dashboards-next/270

Obviously the above dashboard should be updated after this lands /cc @ksiegler1

scholtzan (approved):

PR #324, opened 2021-12-08
Adjust series colours and y-axis scale for opmon.

emtwo:
Some UI tweaks for the generated opmon dashboards

### DIFF
--- a/generator/dashboards/operational_monitoring_dashboard.py
+++ b/generator/dashboards/operational_monitoring_dashboard.py
@@ -39,6 +39,17 @@ class OperationalMonitoringDashboard(Dashboard):
         title = lookml_utils.slug_to_title(name)
         return klass(title, name, "newspaper", namespace, defn["tables"])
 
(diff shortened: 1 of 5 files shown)


generator/dashboards/operational_monitoring_dashboard.py
```diff
@@ -68,13 +79,17 @@ def to_lookml(self, bq_client, data):
...
+                table_defn["branches"], explore
+            )
             for metric in metrics:
                 title = lookml_utils.slug_to_title(metric)
                 kwargs["elements"].append(
                     {
                         "title": title,
                         "metric": metric,
                         "explore": explore,
+                        "series_colours": series_colours,
```
scholtzan, line 92: I think we should be consistent with `color` vs `colour`. Looker seems to be expecting `color`.

```suggestion
                       "series_colors": series_colors,
```

scholtzan (approved):

PR #325, opened 2021-12-09
Disallow rally_markup_fb_pixel_hunt

scholtzan:


whd:
We should have a rule to disallow rally-* and pioneer-* by default.

scholtzan:
Yup, allowing wildcard characters would be good 👍 I'll open an issue for tracking

PR #333, opened 2022-01-04
Fix search alerts table view

scholtzan:


custom-namespaces.yaml
```diff
@@ -73,7 +73,7 @@ search:
       tables:
         - table: mozdata.analysis.desktop_search_alert_records
     desktop_search_alert_latest_daily:
-      type: ping_view
+      type: table_view
```
scholtzan, line 76: The table doesn't have a `submission_date` field

PR #361, opened 2022-02-10
Migrate user_journey

scholtzan:
Migrating `user_journey` namespace which has previously been managed manually

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -604,3 +604,22 @@ regrets_reporter:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.regrets_reporter_derived.regrets_reporter_summary_v1
+messaging_system:
+  glean_app: false
+  pretty_name: Messaging System
+  owners:
+    - mviar@mozilla.com
+    - brennie@mozilla.com
+    - aoprea@mozilla.com
+    - dmosedale@mozilla.com
+    - elee@mozilla.com
+    - emcminn@mozilla.com
+    - pdahiya@mozilla.com
+    - tspurway@mozilla.com
+  spoke: looker-spoke-default
+  views:
+    events_daily:
+      type: table_view
+      tables:
+        - channel: release
+          table: mozdata.messaging_system.events_daily


custom-namespaces.yaml
```diff
@@ -604,3 +604,21 @@ regrets_reporter:
...
+  glean_app: false
+  pretty_name: User Journey
+  owners:
+    - mviar@mozilla.com
+    - brennie@mozilla.com
+    - aoprea@mozilla.com
+    - dmosedale@mozilla.com
+    - elee@mozilla.com
+    - emcminn@mozilla.com
+    - pdahiya@mozilla.com
```
scholtzan, line 617: @lmegviar These are all the users that will have write access to the folder. I looked them up via the People Directory, but feel free to add/remove people as necessary.

custom-namespaces.yaml
```diff
@@ -604,3 +604,21 @@ regrets_reporter:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.regrets_reporter_derived.regrets_reporter_summary_v1
+user_journey:
+  glean_app: false
+  pretty_name: User Journey
```
scholtzan, line 609: @lmegviar you mentioned that your team is now called OMC (what does that stand for?). The folders you see in Looker do not necessarily map 1:1 to separate team but rather to functional groups. You can think of these groups either as products we have (e.g. FF Desktop, Fenix, VPN, ...) or some general aspects we are interested in (e.g. Accessibility, Revenue, Search, User Journey?, ...). These folders (or namespaces as they are called in Looker) will contain dashboards, view, explores that can be used to answer questions to the area of interest.
Calling it `omc` might not be the most descriptive name. What does your team do, maybe there is a better name? What type of questions will the dashboards etc answer?

lmegviar: @scholtzan OMC = Onboarding, Messaging, & Communication. Our team handles onboarding users (such as `about:newtab`) and providing flexible (often remotely configurable) surfaces for communicating with users. These dashboards will primarily be used to answer questions related to messaging experiments powered by Nimbus.

custom-namespaces.yaml
lmegviar: @scholtzan I'd also add our engineering manager at tspurway@mozilla.com.

custom-namespaces.yaml
scholtzan: How about we call it `messaging_system` or `messaging`? If your team starts working on something else, we can always add more namespaces.

PR #423, opened 2022-04-25
delete content.py file, move environment setup to spoke.py file

Marlene-M-Hirose:


generator/spoke.py
```diff
@@ -10,7 +10,7 @@
 import looker_sdk
 import yaml
 
-from .content import setup_env_with_looker_creds
+# from .content import setup_env_with_looker_creds
```
scholtzan, line 13: we can remove this
```suggestion
```

scholtzan (approved):

generator/spoke.py
Marlene-M-Hirose: done

PR #428, opened 2022-04-29
refactor script to use wild cards

Marlene-M-Hirose:


namespaces-disallowlist.yaml
```diff
@@ -10,7 +10,6 @@
...
 - mozphab
 - mozregression
-- rally_core
-- rally_markup_fb_pixel_hunt
-- rally_study_zero_one
-- reference_browser
\ No newline at end of file
+- rally_*
+- reference_browser
+- pioneer-*
```
scholtzan, line 15: These wild-card expressions aren't handled by the code at the moment, so we'll need to support them somewhere around here: https://github.com/mozilla/lookml-generator/blob/ed0d34060b6262ad7730da6d94d742071f13bbc5/generator/namespaces.py#L308-L317

PR #434, opened 2022-05-10
feat: Dockerfile tweaks

kik-kik:
# feat: Dockerfile tweaks

List of changes:
- Added `.vscode` to `.gitignore`
- `.dockerignore` added to exclude some files from the image which I believe should not be needed
- Updated deprecated `MAINTAINER` tag to `LABEL`
- Making sure `no-cache` flags are passed when installing packages
- Condensed Docker commands into fewer commands
- Commented out (for now) 3 OS packages, image builds fine. Could those be needed for anything else?

### DIFF
--- a/.Dockerignore
+++ b/.Dockerignore
@@ -1,0 +1,21 @@
+### A list of files and file patterns to be excluded from the Dockerfile
+
+# 3rd-party configuratiojn
+.circleci
+.github
+.vscode
+.gitignore
+
+# Config files
+.pre-commit-config.yaml
+
+# Docker artifacts
+Dockerfile
+docker-compose.yml
+
+# Python artifacts
+requirements.in
+*.pyc
+
+# Other
+Makefile
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM python:3.8.9-slim
-MAINTAINER Frank Bertsch <frank@mozilla.com>
+
+LABEL maintainer="kignasiak@mozilla.com"
 
 ENV USER_ID="10001"
 ENV GROUP_ID="app"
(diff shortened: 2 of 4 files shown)


.Dockerignore
```diff
@@ -0,0 +1,21 @@
+### A list of files and file patterns to be excluded from the Dockerfile
+
+# 3rd-party configuratiojn
```
scholtzan, line 3:
```suggestion
# 3rd-party configuration
```

Dockerfile
```diff
@@ -9,31 +10,43 @@ RUN groupadd --gid ${USER_ID} ${GROUP_ID} && \
...
-RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-key C99B11DEB97541F0
-RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-key C99B11DEB97541F0
-RUN apt-add-repository https://cli.github.com/packages
-RUN apt update
-RUN apt install -y gh
+RUN apt-get update -qqy \
+    && apt-get install --no-install-recommends -qqy \
+        # python-dev \
+        # build-essential \
+        # curl \
```
scholtzan, line 17: I think we still need those, for example for this: https://github.com/mozilla/lookml-generator/blob/df244e9a545a057776f15d277e63db8f7d9a1172/bin/generate#L235
```suggestion
        python-dev \
        build-essential \
        curl \
```

.Dockerignore
```diff
@@ -0,0 +1,21 @@
+### A list of files and file patterns to be excluded from the Dockerfile
```
relud, line 1: I just tested it, and at least for me this file needs to be `.dockerignore` (all lower case)

relud (dismissed):
r+wc

.Dockerignore
kik-kik: oh no, me and my typos 🙈

.Dockerignore
kik-kik: yes, you are right. I thought I renamed it, but I guess I did not. Should be fixed now.

Dockerfile
kik-kik: do we need `python-dev` and `build-essential` though? Is there any way for us to test this?

scholtzan (approved):

Dockerfile
scholtzan: Turns out they are used for `ssh-keyscan`. Adding them back

PR #440, opened 2022-05-11
Fix ssh-keygen missing in docker

scholtzan:


kik-kik:
Thank you @scholtzan 🙇

Issue #326, opened 2021-12-09
Allow wildcard characters in namespace-disallowlist.yaml


scholtzan:
Follow-up to https://github.com/mozilla/lookml-generator/pull/325#issuecomment-990373833

It should be possible to express something like the following in `namespace-disallowlist.yaml`:

```yaml
- rally-*
- pioneer-*
```

Marlene-M-Hirose:
This has been done: https://github.com/mozilla/lookml-generator/pull/454

PR #454, opened 2022-06-01
add wildcard to rally- and pioneer-

Marlene-M-Hirose:
This refactoring addresses this issue: https://github.com/mozilla/lookml-generator/issues/326

### DIFF
--- a/generator/namespaces.py
+++ b/generator/namespaces.py
@@ -337,7 +337,7 @@ def namespaces(custom_namespaces, generated_sql_uri, app_listings_uri, disallowl
 
     updated_namespaces = {}
     for namespace, _ in namespaces.items():
-        if not disallowed_namespaces_pattern.match(namespace):
+        if not disallowed_namespaces_pattern.fullmatch(namespace):
             if "spoke" not in namespaces[namespace]:
                 namespaces[namespace]["spoke"] = DEFAULT_SPOKE
             if "glean_app" not in namespaces[namespace]:
--- a/generator/namespaces.py
+++ b/generator/namespaces.py
@@ -1,4 +1,5 @@
 """Generate namespaces.yaml."""
+import fnmatch
 import gzip
 import json
 import re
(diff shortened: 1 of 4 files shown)


namespaces-disallowlist.yaml
```diff
@@ -10,8 +10,8 @@
 - mlhackweek_search
 - mozphab
 - mozregression
-- rally_core
-- rally_markup_fb_pixel_hunt
-- rally_study_zero_one
-- reference_browser
 - pine
+- pioneer-*
```
relud, line 14: Looks like namespaces are normally in snake_case, so this probably needs to be:
```suggestion
- pioneer_*
```

relud (changes requested):
this is missing an update to disallowlist handling to support wildcards:

https://github.com/mozilla/lookml-generator/blob/b19c1228b76e8bc218a388b9f2036225b7819037/generator/namespaces.py#L335

generator/namespaces.py
```diff
@@ -329,10 +330,14 @@ def namespaces(custom_namespaces, generated_sql_uri, app_listings_uri, disallowl
...
     disallowed_namespaces = yaml.safe_load(disallowlist.read()) or {}
+    disallowed_regex = [
+        fnmatch.translate(namespace) for namespace in disallowed_namespaces
+    ]
+    disallowed_namespaces_pattern = re.compile("|".join(disallowed_regex))
 
     updated_namespaces = {}
     for namespace, _ in namespaces.items():
-        if namespace not in disallowed_namespaces:
+        if not disallowed_namespaces_pattern.match(namespace):
```
relud, line 340: this needs to use fullmatch to ensure it's matching the entire string:
```suggestion
        if not disallowed_namespaces_pattern.fullmatch(namespace):
```

note: `re.Pattern.match` "Matches zero or more characters at the beginning of the string", and `fnmatch.translate` will suffix each pattern with `\Z` (end of input), so this isn't required, but it's still safer to use `fullmatch` in case someone changes how `disallowed_namespaces_pattern` is generated.

relud (commented):
`tests/test_namespaces.py` needs to be updated to test that this works

tests/test_namespaces.py
```diff
@@ -101,7 +111,7 @@ def namespace_disallowlist(tmp_path):
         dedent(
             """
             ---
-            disallowed
+            - disallowed*
```
relud, line 114: I think it would be good to still test that non-wildcards work:
```suggestion
            - disallowed
            - disallowed_*
```

relud (approved):
r+wc

PR #456, opened 2022-06-02
Added acoustic_contacts_current_snapshot view (+ rename)

kik-kik:
renamed a view to make it explicit that it is change events and added view for as of today state (build on top of a dedup table for acoustic contacts)

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -773,10 +773,14 @@ marketing:
   owners:
     - kignasiak@mozilla.com
   views:
-    acoustic_contacts:
+    acoustic_contact_change_events:
       type: table_view
       tables:
         - table: moz-fx-data-marketing-prod.acoustic.contact_v1
+    acoustic_contacts_current_snapshot:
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -773,10 +773,14 @@ marketing:
   owners:
     - kignasiak@mozilla.com
   views:
-    acoustic_contacts:
+    acoustic_contact_change_events:
```
lucia-vargas-a, line 776: Would it be better to create a new one and then remove the previous one, to avoid conflicts with a non existing namespace in Looker?

kik-kik: This namespace is not really used yet, and I have another PR to add explores for those views which I have modified to compensate for this change. I just thought since usage is limited not then I might as well use the opportunity to make names a bit more explicit

lucia-vargas-a: Sound good

lucia-vargas-a (approved):

PR #490, opened 2022-07-21
Switch medatada daily view to the new updated table in BQ

chelseatroy:
We updated the table schema!

Now this view needs to pull from this table.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -97,7 +97,7 @@ search_private:
     search_term_sanitization_job_metadata_daily:
       type: table_view
       tables:
-        - table: mozdata.search_terms_unsanitized_analysis.prototype_sanitation_job_metadata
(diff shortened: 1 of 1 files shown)


scholtzan (commented):
Does the table already exist? When I search for it in BigQuery I can't find it

whd:
I suspect the intent here is to transition to the production managed table `moz-fx-data-shared-prod.search_terms_derived.sanitization_job_metadata_v2`: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/search_terms_derived/sanitization_job_metadata_v2/init.sql#L2

chelseatroy:
1. Ah, yes, that one doesn't exist
2. Yes, the one @whd mentioned should exist and is the one I meant :)

I'll fix this now!

scholtzan (approved):

PR #491, opened 2022-07-22
DO-827 Create firefox iOS topsites impression event ping view and explore

ANich:
Similar to https://github.com/mozilla/lookml-generator/pull/385  -- RE https://mozilla-hub.atlassian.net/browse/DO-827


ANich:
Follow-up I'm not sure why the default created Ping Explore doesn't include events. cc @scholtzan

Marlene-M-Hirose (approved):

PR #506, opened 2022-08-25
Set document_id as primary_key in event views when available

scholtzan:
This PR fixes the `cannot be calculated because of a one_to_many or many_to_many join` errors in the event explores.

These docs were helpful in figuring out the problem: https://cloud.google.com/looker/docs/best-practices/how-to-troubleshoot-when-measures-dont-come-through-a-join

To enable measures to carry through joins (such as client count and event count), we need to define a primary key in the involved views. `document_id` is the natural choice here as primary key for the event views.

I pushed the changes to a branch in looker-hub to test them and the join worked just as expected: https://github.com/mozilla/looker-hub/tree/ascholtz-events-join-test-1

cc @skahmann3

### DIFF
--- a/generator/views/events_view.py
+++ b/generator/views/events_view.py
@@ -73,6 +73,14 @@ class EventsView(View):
         )
         view_defn["measures"] = self.get_measures(dimensions)
 
+        # set document_id as primary key if it exists in the underlying table
+        # this will allow one_to_many joins
+        document_id_field = self.get_document_id(dimensions, "events")
+        if document_id_field is not None:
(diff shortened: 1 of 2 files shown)


scholtzan:
Another thing to investigate, and maybe it's worth opening an issue for this, is whether `document_id` dimensions should *always* be set a primary key. This change only sets them as such for the event views, but we'll be doing joins on other explores/views which might fail with the same issue

Marlene-M-Hirose (approved):

PR #511, opened 2022-08-30
Remove requirement to tag deploys. Use -latest instead

scholtzan:
See discussion in https://mozilla.slack.com/archives/C01E8GDG80N/p1661854246465379

This should remove the requirement of manually tagging releases. Changes will need to be made to Airflow to use the `-latest` image


fbertsch (approved):
LGTM, but do we need a dupe step with a tag of `${CIRCLE_SHA1}`? Then we can simply change the Airflow variable to be `latest` (for now) and switch to a commit if something goes wrong

scholtzan:
Good point

kik-kik (approved):

akkomar (commented):
+1 for tagging with commit hash.

.circleci/config.yml
```diff
@@ -178,9 +160,18 @@ workflows:
...
+      
+      - gcp-gcr/add-image-tag:
+          name: Add commit hash as docker image tag
+          context: data-eng-airflow-gcr
+          requires:
+            - Build and push lookml-generator
+          image: lookml-generator
+          source-tag: latest
+          target-tag: $CIRCLE_SHA1
+          filters:
```
scholtzan, line 175: Alright, this should add the commit hash as a tag

PR #516, opened 2022-09-07
Bug(DS-2132): FxA accounts dashboard is broken client id not found

kik-kik:
# bug(DS-2132): FxA accounts dashboard is broken client id not found

Seems the change resulted in expected changes:
https://github.com/mozilla/looker-hub/blob/kik-generation-test-1/firefox_accounts/views/growth_accounting.view.lkml#L103-L108

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -334,6 +334,7 @@ firefox_accounts:
           table: mozdata.firefox_accounts.events_daily
     growth_accounting:
       type: growth_accounting_view
+      primary_key_field: user_id
       tables:
         - table: mozdata.firefox_accounts.fxa_users_last_seen
     fxa_first_seen_table:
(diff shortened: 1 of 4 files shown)


custom-namespaces.yaml
```diff
@@ -334,6 +334,7 @@ firefox_accounts:
           table: mozdata.firefox_accounts.events_daily
     growth_accounting:
       type: growth_accounting_view
+      primary_key_field: user_id
```
fbertsch, line 337: nit: this field isn't a primary key, it has to be combined with the `submission_date` to qualify as that. Maybe call it `identifier_field`?

kik-kik: fair enough, renamed the field.

scholtzan (approved):
looks good to me

kik-kik:
Finally managed to generate the dev branch and check it in Looker. Seems this fixes the issue identified in the ticket.

PR #517, opened 2022-09-08
default source branch set to main in generate script

kik-kik:
I'm not sure if this is a mistake or intended. I suspect this is why my generated branch results with 400 commits behind main in looker-hub repo.

However, now trying to run make run command seems to result in the following kinds of errors:
```
Error: lookml-generator modified <view_path>.view.lkml
```

So perhaps the use of `base` branch here is correct?


kik-kik:
closing as base branch is correct.

PR #537, opened 2022-10-12
Add date filter to opmon dashboards

relud:
fixes mozilla/opmon#14


generator/views/operational_monitoring_view.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, namespace: str, name: str, tables: List[Dict[str, Any]]):
                 "name": xaxis,
                 "type": "date",
                 "sql": xaxis_to_sql_mapping[xaxis],
+                "datatype": "date",
```
relud, line 39: without this the filter will try to compare a date to a timestamp, which bigquery does not allow.

scholtzan (approved):

PR #545, opened 2022-10-19
Reference url metrics under their other name (`url2`) in generated views

### LINKED ISSUE #539: Expose `url` types in Glean Ping views
The `url` type was added to the Glean client and is being used to send data for MDN, but this codebase doesn't propagate those fields to the Looker views. This ticket is for exposing them.

It should start with exposing them as `ALLOWED_TYPES` here (that might be everything, but also add a test).

badboy:
Fixes #539

### DIFF
--- a/generator/views/glean_ping_view.py
+++ b/generator/views/glean_ping_view.py
@@ -29,6 +29,15 @@ ALLOWED_TYPES = DISTRIBUTION_TYPES | {
     "rate",
     "timespan",
     "uuid",
+    "url",
+}
+
+# Bug 1737656 - some metric types are exposed under different names
+# We need to map to the new name when building dimensions.
+RENAMED_METRIC_TYPES = {
+    "jwe": "jwe2",
+    "text": "text2",
+    "url": "url2",
 }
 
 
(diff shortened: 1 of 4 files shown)


generator/views/glean_ping_view.py
```diff
@@ -29,6 +29,14 @@
     "rate",
     "timespan",
     "uuid",
+    "url",
+}
+
+# Some metric types are exposed under different names (due to bugs).
```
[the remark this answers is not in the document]
fbertsch: Oops, you can ignore this. I see the bug number below

generator/views/glean_ping_view.py
```diff
@@ -29,6 +29,16 @@
     "rate",
     "timespan",
     "uuid",
+    "url",
```
fbertsch, line 32: looks like we need `labeled_rate` and `text` here

badboy: `labeled_rate` does not really exist (which means I should actually remove it from the other list, too)
`text` exist, but not currently used (afaik) and in general category 3 or 4. Not sure if we want to default-expose that.

badboy:
ok, now I think it is ready

PR #559, opened 2022-11-08
Support multiple client count views in a single namespace

scholtzan:
The `firefox_desktop` namespace now has multiple client count explores. However, when generating the views they are currently always named `client_counts`. So the same view gets overridden. Instead we want to be able to generate multiple client count views and use their configured names.

cc @rebecca-burwei

### DIFF
--- a/generator/views/client_counts_view.py
+++ b/generator/views/client_counts_view.py
@@ -67,9 +67,14 @@ class ClientCountsView(View):
         }
     ]
 
-    def __init__(self, namespace: str, tables: List[Dict[str, str]]):
+    def __init__(
+        self,
+        namespace: str,
+        tables: List[Dict[str, str]],
+        name: str = "client_counts",
+    ):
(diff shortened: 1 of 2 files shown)


generator/views/client_counts_view.py
```diff
@@ -67,9 +67,14 @@ class ClientCountsView(View):
...
-    def __init__(self, namespace: str, tables: List[Dict[str, str]]):
+    def __init__(
+        self,
+        namespace: str,
+        tables: List[Dict[str, str]],
+        name: str = "client_counts",
+    ):
         """Get an instance of a ClientCountsView."""
-        super().__init__(namespace, "client_counts", ClientCountsView.type, tables)
+        super().__init__(namespace, name, ClientCountsView.type, tables)
```
scholtzan, line 77: this change shouldn't change existing views since all of them are named `client_counts` anyway

Marlene-M-Hirose (approved):

PR #562, opened 2022-11-19
Added usage and table views to monitoring namesapce

wwyc:
Added usage and table views to monitoring namesapce

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -425,7 +425,7 @@ revenue_private:
     revenue_budget_2022:
       type: table_view
       tables:
-        - table: mozdata.revenue_cat3_analysis.budget_complete_2022       
+        - table: mozdata.revenue_cat3_analysis.budget_complete_2022
 firefox_ios:
   owners:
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -680,12 +680,29 @@ monitoring:
...
+    - wichan@mozilla.com
   pretty_name: Monitoring (Data Pipeline)
   views:
     average_ping_sizes:
       type: ping_view
       tables:
         - table: mozdata.monitoring.average_ping_sizes
+    bigquery_tables_inventory:
+      tables:
+        - table: moz-fx-data-shared-prod.monitoring.bigquery_tables_inventory
```
ANich, line 692: Why not:
```suggestion
        - table: mozdata.monitoring.bigquery_tables_inventory
```
here and below?

ANich (approved):

custom-namespaces.yaml
wwyc: This is to align with other monitoring datasets to be in the same project.

PR #576, opened 2022-12-12
Fix Airflow tables

scholtzan:
To unblock lookml-generator from failing


kik-kik (approved):
Oops, thanks @scholtzan ! 🙇

PR #578, opened 2022-12-13
add new adM forecasts table

skahmann3:


custom-namespaces.yaml
```diff
@@ -162,6 +162,10 @@ contextual_services:
     - rburwei@mozilla.com
   spoke: looker-spoke-private
   views:
+    final_adM_forecasts:
+      type: ping_view
```
scholtzan, line 166:
```suggestion
      type: table_view
```
Not sure what the structure of the spreadsheet is, but ping_views are usually used for ping tables and require a submission_date, client_id, and document_id column

custom-namespaces.yaml
```diff
@@ -213,6 +217,10 @@ contextual_services:
           table: mozdata.org_mozilla_ios_fennec.topsites_impression
       type: ping_view
   explores:
+    final_adM_forecasts:
+      type: ping_explore
+      views:
+        base_view: final_adM_forecasts
```
scholtzan, line 223:
```suggestion
```
If the spreadsheet is a `table_view` then lookml-generator currently does not generate explores for those. But you can create a custom one once the view has been deployed

custom-namespaces.yaml
```diff
@@ -162,6 +162,10 @@ contextual_services:
     - rburwei@mozilla.com
   spoke: looker-spoke-private
   views:
+    final_adM_forecasts:
```
scholtzan, line 165: nit. usually we just use lowercase identifiers
```suggestion
    final_adm_forecasts:
```

scholtzan (approved):

PR #579, opened 2022-12-15
Add Glean Events explore for Firefox Desktop

Iinh:
More context: https://mozilla-hub.atlassian.net/browse/LOOK-129


ANich (approved):
This LGTM relative to other implementations in `custom-namespaces` though I'm not 100% sure whether glean generation is different from legacy generation

PR #608, opened 2023-02-02
add fivetran costs to looker

lelilia:
I want to add my new view to looker, so please let me know if that is how to do it :)

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1053,11 +1053,11 @@ relay:
       type: table_view
       tables:
         - table: mozdata.relay.active_subscription_ids
-fivetran_costs:
+fivetran:
   glean_app: false
   owners:
     - lschiestl@mozilla.com
-  pretty_name: Fivetran Costs
+  pretty_name: Fivetran
   views:
     monthly_connector_costs:
       type: table_view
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -1053,3 +1053,13 @@ relay:
       type: table_view
       tables:
         - table: mozdata.relay.active_subscription_ids
+fivetran_costs:
```
scholtzan, line 1056:
```suggestion
fivetran:
```
should we just call it `fivetran`? We might want to add more views/explores that are related to Fivetran but not related to cost specifically in the future?

custom-namespaces.yaml
```diff
@@ -1053,3 +1053,13 @@ relay:
...
+  glean_app: false
+  owners:
+    - lschiestl@mozilla.com
+  pretty_name: Fivetran Costs
```
scholtzan, line 1060:
```suggestion
  pretty_name: Fivetran
```

custom-namespaces.yaml
lelilia: good point ... but the dataset is now named fivetran_costs ... should that be renamed as well or could I put multiple datasets into this namespace?

PR #615, opened 2023-02-13
Generate datagroups

ANich:
Implementation of the [datagroup generation proposal](https://docs.google.com/document/d/1dPwa5KZnKHzYj3hULHy9unczCXugwr6rZHW1gveQ2fc/edit#heading=h.5x0d5h95i329), see also [DENG-676](https://mozilla-hub.atlassian.net/browse/DENG-676).

Split into a small refactor to use the generated-sql function, then datagroup generation with tests.

### DIFF
--- a/generator/lookml.py
+++ b/generator/lookml.py
@@ -12,6 +12,7 @@ from .dashboards import DASHBOARD_TYPES
 from .explores import EXPLORE_TYPES
 from .namespaces import _get_glean_apps
 from .views import VIEW_TYPES, View, ViewDict
+from .views.datagroups import generate_datagroups
 
 FILE_HEADER = """
 # *Do not manually modify this file*
@@ -119,6 +120,9 @@ def _lookml(namespaces, glean_apps, target_dir):
         v1_name: Optional[str] = v1_mapping.get(namespace)
         for view_path in _generate_views(client, view_dir, views, v1_name):
             logging.info(f"    ...Generating {view_path}")
+
+        logging.info("  Generating datagroups")
+        generate_datagroups(views, target, namespace, client)
 
         explore_dir = target / namespace / "explores"
         explore_dir.mkdir(parents=True, exist_ok=True)
--- a/generator/views/datagroups.py
+++ b/generator/views/datagroups.py
@@ -1,0 +1,162 @@
+"""Generate datagroup lkml files for each namespace."""
+
+import logging
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, List, Optional
+
+import lkml
+from google.api_core.exceptions import NotFound
+from google.cloud import bigquery
+
+from generator.namespaces import DEFAULT_GENERATED_SQL_URI
+from generator.views import TableView, View, lookml_utils
+from generator.views.lookml_utils import BQViewReferenceMap
+
+DEFAULT_MAX_CACHE_AGE = "24 hours"
+DEFAULT_INTERVAL_TRIGGER = "6 hours"
+
+# Note: INFORMATION_SCHEMA.PARTITIONS has a row with a `last_modified_time` value even for non-partitioned tables.
+SQL_TRIGGER_TEMPLATE = """
+    SELECT MAX(last_modified_time)
+    FROM `{project_id}`.{dataset_id}.INFORMATION_SCHEMA.PARTITIONS
+    WHERE table_name = '{table_id}'
+"""
+
+FILE_HEADER = """# *Do not manually modify this file*
+
+# This file has been generated via https://github.com/mozilla/lookml-generator
+
+# Using a datagroup in an Explore: https://cloud.google.com/looker/docs/reference/param-explore-persist-with
+# Using a datagroup in a derived table: https://cloud.google.com/looker/docs/reference/param-view-datagroup-trigger
+
+"""
+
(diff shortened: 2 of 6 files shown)


generator/views/datagroups.py
```diff
@@ -0,0 +1,162 @@
...
+
+def generate_datagroups(
+    views: List[View], target_dir: Path, namespace: str, client: bigquery.Client
+) -> None:
+    """Generate and write a datagroups.lkml file to the namespace folder."""
+    datagroups_lkml_path = target_dir / namespace / "datagroups.lkml"
+
+    # To map views to their underlying tables:
+    dataset_view_map = lookml_utils.get_bigquery_view_reference_map(
+        DEFAULT_GENERATED_SQL_URI
```
ANich, line 152: Elsewhere this is a CLI parameter but I've just used the default here for simplicity. If there's a reason not do this I can pass this in as a param in the upstream `lookml` function

generator/lookml.py
```diff
@@ -120,6 +121,9 @@ def _lookml(namespaces, glean_apps, target_dir):
         for view_path in _generate_views(client, view_dir, views, v1_name):
             logging.info(f"    ...Generating {view_path}")
 
+        logging.info("  Generating datagroups")
```
kik-kik, line 124: Should we use f-string settings for managing the padding? Then we could globally define different indentation levels?
This is what I mean: https://stackoverflow.com/a/5676884

generator/views/datagroups.py
```diff
@@ -0,0 +1,162 @@
...
+        ) from e
+
+    if "TABLE" == bq_table.table_type:
+        datagroup: Optional[Datagroup] = _get_datagroup_from_bigquery_table(bq_table)
+        return str(datagroup)
+    elif "VIEW" == bq_table.table_type:
+        datagroup = _get_datagroup_from_bigquery_view(
+            bq_table, client, dataset_view_map
+        )
+        if datagroup is not None:
```
kik-kik, line 138:
```suggestion
        if not datagroup:
```

tests/test_datagroups.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+@patch("google.cloud.bigquery.Table")
+@patch("google.cloud.bigquery.Table")
+@patch("google.cloud.bigquery.Client")
+@patch("generator.views.lookml_utils.get_bigquery_view_reference_map")
+def test_generates_datagroups(reference_map_mock, client, table_1, table_2, runner):
+    expected = """# *Do not manually modify this file*
+
+# This file has been generated via https://github.com/mozilla/lookml-generator
```
kik-kik, line 33: Love the docs for the test

generator/views/lookml_utils.py
```diff
@@ -209,3 +214,29 @@ def get_distinct_vals(bq_client: bigquery.Client, table: str, column: str):
...
+# Map from view to qualified references {view: [[project, dataset, table],]}
+BQViewReferenceMap = Dict[str, List[List[str]]]
+
+
+def get_bigquery_view_reference_map(
+    generated_sql_uri: str,
+) -> Dict[str, BQViewReferenceMap]:
+    """Get a mapping from BigQuery datasets to views with references."""
+    with urllib.request.urlopen(generated_sql_uri) as f:
+        tarbytes = BytesIO(f.read())
```
kik-kik, line 228:
```suggestion
        tarbytes = BytesIO(f.read())
        
```

generator/views/datagroups.py
relud: as i understand it, `is not None` is actually the python idiomatic way to handle this here.

generator/views/lookml_utils.py
```diff
@@ -209,3 +214,29 @@ def get_distinct_vals(bq_client: bigquery.Client, table: str, column: str):
 def slug_to_title(slug):
     """Convert a slug to title case."""
     return slug.replace("_", " ").title()
+
+
+# Map from view to qualified references {view: [[project, dataset, table],]}
+BQViewReferenceMap = Dict[str, List[List[str]]]
```
relud, line 220: it looks like all the uses of BQViewReferenceMap are in `Dict[str, BQViewReferenceMap]`, why not have the type include that outer dict? i.e.:
```suggestion
BQViewReferenceMap = Dict[str, Dict[str, List[List[str]]]]
```

generator/views/datagroups.py
ANich: Yup, according to the Google style guide anyway: https://google.github.io/styleguide/pyguide.html#2144-decision (when explicitly checking for `None` which is the intention of `_get_datagroup_from_bigquery_view`

relud (approved):
r+ with or without my comments

generator/views/datagroups.py
kik-kik: my bad, thanks for clarifying.

generator/lookml.py
ANich: Yeah I don't see why not, at the very least we could replace some of these with \t (in a follow-up 😆)

generator/views/lookml_utils.py
ANich: Yeah this makes sense, I'll update along with the description comment

PR #617, opened 2023-02-16
Avoid repeating the same table multiple times for a view (DENG-690)

sean-rose:
The same table was being repeated for a view if the Glean app had multiple release channels (due to the [special case logic which overrides the dataset for release channels](https://github.com/mozilla/lookml-generator/blob/09cdf4c404c07c4c5dc0ee8508646bce85f8e1ef/generator/namespaces.py#L202-L220)), which the `mozilla_vpn` app currently does after https://github.com/mozilla/probe-scraper/pull/552.  This resulted in unnecessary `channel` parameters being added to all the affected views.

This will hopefully resolve the errors in the VPN Looker views related to `channel` parameters/fields in [DENG-690](https://mozilla-hub.atlassian.net/browse/DENG-690) "_Various errors in looker-spoke-default_".

However, it doesn't address the fundamental complications involved:
- Some apps like `mozilla_vpn` will have multiple release channels, which should probably be handled in a better way.
- Some datasets like `mozilla_vpn` contain a mix of both Glean tables/views and non-Glean tables/views, but the current LookML generation logic just has a `glean_app` boolean setting at the namespace level.  We may need more granular settings for marking specific tables/views as Glean/non-Glean, or should perhaps avoid having a mix of Glean & non-Glean tables/views in the same dataset.

### DIFF
--- a/generator/views/ping_view.py
+++ b/generator/views/ping_view.py
@@ -30,7 +30,7 @@ class PingView(View):
         if (klass.allow_glean and not is_glean) or (not klass.allow_glean and is_glean):
             return
 
-        views = defaultdict(list)
+        view_tables: Dict[str, Dict[str, Dict[str, str]]] = defaultdict(dict)
         for channel in channels:
             dataset = channel["dataset"]
 
@@ -38,8 +38,8 @@ class PingView(View):
                 if view_id in OMIT_VIEWS:
                     continue
 
-                table: Dict[str, str] = {"table": f"mozdata.{dataset}.{view_id}"}
-
+                table_id = f"mozdata.{dataset}.{view_id}"
+                table: Dict[str, str] = {"table": table_id}
                 if channel.get("channel") is not None:
                     table["channel"] = channel["channel"]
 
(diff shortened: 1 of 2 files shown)


scholtzan (approved):
Would using the unioned Glean app ping views help here? (not as part of this PR, but in general to work around some of the issues mentioned)

sean-rose:
> Would using the unioned Glean app ping views help here? (not as part of this PR, but in general to work around some of the issues mentioned)

The unioned Glean app ping views do already help, because without them the result would be that the pings from additional release channels wouldn't be accessible via these generated explores/views.  Although I also just realized that now with the unioned Glean app ping views the existing `channel` parameter switching doesn't give a way to only get release channel data, as it's pointing to the unioned view in that case.

As it is, one thing that's lost is the ability to use a `channel` parameter to optionally switch between multiple release channels like you can between release/beta/nightly channels.  You can instead filter on the `normalized_app_id` column to do that, but it's not quite the same as you're filtering the unioned view rather than querying a channel-specific view directly and scanning less data.

However, if we do make it so `channel` parameters work for multiple release channels, we'll still hit the complication of `channel` parameters getting added to all non-Glean tables/views that may exist in a Glean app's dataset, which wouldn't be ideal.  Besides the idea of table/view-specific Glean config settings, another idea would be to have the LookML generation logic figure out what Glean ping tables/views should exist (e.g. by reading probe-scraper data) and only treat those as Glean tables/views rather than assuming all tables/views in the dataset are from Glean.

PR #620, opened 2023-02-21
Only generate one datagroup per-table

ANich:
Some views reference the same table. For example `mozdata.org_mozilla_ios_klar.baseline_clients_daily` and `mozdata.org_mozilla_ios_klar.baseline_clients_first_seen` are both views to `moz-fx-data-shared-prod.org_mozilla_ios_klar_derived.baseline_clients_daily_v1`.

We only need a single datagroup generated in cases like this.

### DIFF
--- a/tests/test_datagroups.py
+++ b/tests/test_datagroups.py
@@ -35,22 +35,22 @@ def test_generates_datagroups(reference_map_mock, client, table_1, table_2, runn
 # Using a datagroup in an Explore: https://cloud.google.com/looker/docs/reference/param-explore-persist-with
 # Using a datagroup in a derived table: https://cloud.google.com/looker/docs/reference/param-view-datagroup-trigger
 
-datagroup: test_table_last_updated {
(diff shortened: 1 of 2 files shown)


tests/test_datagroups.py
```diff
@@ -35,22 +35,22 @@ def test_generates_datagroups(reference_map_mock, client, table_1, table_2, runn
 # Using a datagroup in an Explore: https://cloud.google.com/looker/docs/reference/param-explore-persist-with
 # Using a datagroup in a derived table: https://cloud.google.com/looker/docs/reference/param-view-datagroup-trigger
 
-datagroup: test_table_last_updated {
-  label: "Test Table Last Updated"
+datagroup: test_table_2_last_updated {
+  label: "Test Table 2 Last Updated"
```
ANich, line 39: This is all related to sort-order

kik-kik (approved):

PR #636, opened 2023-03-09
Remove some apps from disallowlist

fbertsch:
Per a discussion in slack with @Dexterp37, these are in-use and would be beneficial to have in Looker.

### DIFF
--- a/namespaces-disallowlist.yaml
+++ b/namespaces-disallowlist.yaml
@@ -6,7 +6,6 @@
 - firefox_reality_pc
 - lockwise_android
 - lockwise_ios
-- mach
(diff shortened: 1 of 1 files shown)


namespaces-disallowlist.yaml
```diff
@@ -6,7 +6,6 @@
 - firefox_reality_pc
 - lockwise_android
 - lockwise_ios
-- mach
 - mlhackweek_search
 - mozphab
 - mozregression
```
fbertsch, line 11:
```suggestion
```

scholtzan (approved):

PR #638, opened 2023-03-13
feat: updated datasets for dim / data_monitoring datasets and adding explores

kik-kik:
# feat: updated datasets for dim / data_monitoring datasets

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1028,15 +1028,15 @@ data_integrity_monitoring:
   views:
     run_history:
       tables:
-        - table: data-monitoring-dev.dim.dim_run_history
+        - table: data-monitoring-dev.dim_external.run_history
       type: table_view
     run_processing_info:
       tables:
-        - table: data-monitoring-dev.dim.dim_run_processing_info
+        - table: data-monitoring-dev.dim_external.run_processing_info
       type: table_view
     muted_alerts:
       tables:
-        - table: data-monitoring-dev.dim.dim_muted_alerts
+        - table: data-monitoring-dev.dim_external.muted_alerts
       type: table_view
 relay:
   glean_app: false


custom-namespaces.yaml
```diff
@@ -1028,16 +1028,29 @@ data_integrity_monitoring:
...
+        - table: data-monitoring-dev.dim_external.run_processing_info
       type: table_view
     muted_alerts:
       tables:
-        - table: data-monitoring-dev.dim.dim_muted_alerts
+        - table: data-monitoring-dev.dim_external.muted_alerts
       type: table_view
+  explores:
+    run_history:
+      type: explore
```
fbertsch, line 1043: Is this a valid explore type?

custom-namespaces.yaml
```diff
@@ -1028,16 +1028,29 @@ data_integrity_monitoring:
   views:
     run_history:
       tables:
-        - table: data-monitoring-dev.dim.dim_run_history
+        - table: data-monitoring-dev.dim_external.run_history
       type: table_view
     run_processing_info:
       tables:
-        - table: data-monitoring-dev.dim.dim_run_processing_info
+        - table: data-monitoring-dev.dim_external.run_processing_info
```
lucia-vargas-a, line 1035: LGTM.
You may want to use the standard for datasets that we use in moz-fx-data-shared-prod:
e.g. dim_derived (to store tables) > dim (to store views)

lucia-vargas-a (approved):

custom-namespaces.yaml
kik-kik: It appears that the modules defined here:
https://github.com/mozilla/lookml-generator/tree/main/generator/explores

Correspond to explore types? Could be wrong here.

custom-namespaces.yaml
kik-kik: hmm, I thought about it but `_derived` would also not be correct in this case as this table is populated by the application itself. Could have always called it something like `dim_operational`, but we can always do that if we ever decide to invest more time into it.

custom-namespaces.yaml
kik-kik: I can try running generator after merge and worse case scenario remove the explores here and add them in looker-spoke

kik-kik: I see, thanks for pointing this out. I'll just take it out from here and add it to looker-spoke instead.

fbertsch (approved):

PR #648, opened 2023-03-31
Bug 1823627 - Use a channel filter instead of a parameter switching out the table

badboy:
The parameter switched out the underlying table to be queried. However we now have UNIONed tables for every variant of an application, making the first table contain data from all channels. Instead we now filter by `normalized_channel` and enforce that filter in explores.

### DIFF
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -145,17 +145,13 @@ class Explore:
         channel_params = [
             param
             for _view_defn in self.get_view_lookml(view)["views"]
-            for param in _view_defn.get("parameters", [])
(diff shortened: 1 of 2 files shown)


badboy:
I tested the generation locally (using `namespaces.yaml` from looker-hub trimmed down to Fenix) and in Looker Developer mode by creating a new view/explore with this filter instead of a parameter.

badboy:
While trying to fix the tests I noticed we do make some assumptions around the tables and channels.
And at least for some derived views I'm not sure the assumptions are guaranteed to hold, so I need to look into that further first.

PR #683, opened 2023-06-05
Fix suggestion filter: Only take existing values

badboy:


generator/views/ping_view.py
```diff
@@ -93,7 +93,11 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
         )
 
         # Round-tripping through a dict to get an ordered deduped list.
-        suggestions = list(dict.fromkeys(_table["channel"] for _table in self.tables))
+        suggestions = list(
+            dict.fromkeys(
+                _table.get("channel") for _table in self.tables if "channel" in _table
```
kik-kik, line 98: Do you want to exclude things here completely that do not have `channel`?

badboy: Yes. If there's no channel defined, we can't filter by it.

My reasoning for doing it like that was:
If there's multiple tables then probably none of them has `channel` defined, which means `suggestions` will be empty and the whole filter logic will be skipped.

If there would be an app with multiple tables, but only some of them have no channel and others do, this logic will still do the right thing and filter by that defined channel.
At least this wouldn't explode.

kik-kik: In that case instead of `_table.get("channel")` Should we do `_table["channel"]` since the key should always exist here?

badboy: Ah I guess we could

kik-kik (approved):

PR #685, opened 2023-06-05
RS 733 - Create public ctx srvcs namespace

alekhyamoz:
Relates to https://mozilla-hub.atlassian.net/browse/RS-733. 
1. Rename `contextual_services` to `contextual_services_private` (spoke remains `looker-spoke-private`)
2. Add new namespace `contextual_services` which is defaulted to public.
Note - the `pretty_name` will remain `Contextual Services` for both namespaces
3. Add the fill rate `mozdata view` to the public namespace.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -155,7 +155,7 @@ search:
       type: ping_explore
       views:
         base_view: desktop_search_alert_records
-contextual_services:
+contextual_services_private:
   glean_app: false
   connection: bigquery-oauth
   pretty_name: Contextual Services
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -268,6 +268,19 @@ contextual_services:
       type: ping_explore
       views:
         base_view: topsites_impression_firefox_ios
+contextual_services:
+  glean_app: false
+  connection: bigquery-oauth
```
scholtzan, line 273: is oauth necessary for these tables?

Users will need to authenticate to BigQuery for any dashboard tile that uses them. It looks like the one table listed here is accessible to Mozilla. So we could remove `connection` here so that any access is done over the Looker service account (and users don't need to authenticate).

```suggestion
```

scholtzan (approved):
lgtm, let's see if it breaks stuff

skahmann3:
Sweet, so users will see "Contextual Services" when surfing for an Explore, and then data will render or not based on user's permissions?

alekhyamoz:
@skahmann3 please let me know if we need to move any other tables between `public` and `private` or we can do it in iterations after confirming that this PR would not break anything. :)

skahmann3 (approved):
Thank you both for helping with this!! Let’s try it!

PR #686, opened 2023-06-05
Allow `table_view` measures to be defined in `custom-namespaces.yaml`

sean-rose:
Example:
```yaml
foo_namespace:
  views:
    bar_table:
      type: table_view
      tables:
        - table: ...
      measures:
        count:
          type: count
        client_count:
          type: count_distinct
          sql: client_id
```

### DIFF
--- a/generator/views/view.py
+++ b/generator/views/view.py
@@ -8,6 +8,7 @@ from click import ClickException
 OMIT_VIEWS: Set[str] = set()
 
 
+# TODO: Once we upgrade to Python 3.11 mark just `measures` as non-required, not all keys.
 class ViewDict(TypedDict, total=False):
     """Represent a view definition."""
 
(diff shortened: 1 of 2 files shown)


generator/views/view.py
```diff
@@ -8,11 +8,12 @@
 OMIT_VIEWS: Set[str] = set()
 
 
-class ViewDict(TypedDict):
+class ViewDict(TypedDict, total=False):
```
sean-rose, line 11: Marking all keys as non-required isn't ideal, but I couldn't figure out another way to satisfy mypy's type checks at the moment.

In Python 3.11 `TypedDict` supports marking individual keys as non-required, so whenever we upgrade we can adjust this.

relud: this should be a comment in the file, so we know to fix it later.

PR #687, opened 2023-06-05
Add count measures to SubPlat table views (DS-2934)

sean-rose:
## [DS-2934](https://mozilla-hub.atlassian.net/browse/DS-2934): Make current SubPlat ETLs available as Looker explores

This is a follow-up to #684.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1131,14 +1131,23 @@ subscription_platform:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.subscription_platform_derived.apple_subscriptions_v1
+      measures:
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -1131,14 +1131,23 @@ subscription_platform:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.subscription_platform_derived.apple_subscriptions_v1
+      measures:
+        count:
+          type: count
```
sean-rose, line 1136: Defining measures like this is a new feature for table views just added in #686.

akkomar (approved):

PR #689, opened 2023-06-08
Move includes in spoke model generation to footer comment.

ANich:
By default, spoke-default PRs (e.g. https://github.com/mozilla/looker-spoke-default/pull/563) that create models auto-generate includes to a number of directories (as well as create empty directories). It seems like pointing to the empty created directories no longer prevents LookML warnings. One way to solve this is by manually editing the PR to comment out or remove the includes that won't have content. 

This PR automatically generates the includes as comments so merging by default won't break the production spokes.

### DIFF
--- a/generator/spoke.py
+++ b/generator/spoke.py
@@ -83,7 +83,7 @@ def generate_model(
 
     path = spoke_path / name / f"{name}.model.lkml"
     # lkml.dump may return None, in which case write an empty file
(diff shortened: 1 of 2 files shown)


ANich:
I'm not sure whether this will open new pull requests on the next generator run or only apply to future future added models (namespaces) 🤔

scholtzan (approved):

PR #692, opened 2023-06-13
feat(DENG-1017): added an events view and events explore for fxa_all_events

kik-kik:
# feat(DENG-1017): added an events view and events explore for fxa_all_events

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -338,10 +338,10 @@ firefox_accounts:
     - loines@mozilla.com
   pretty_name: Firefox Accounts
   views:
-    fxa_all_events:
+    all_events:
       type: events_view
       tables:
-        - events_table_view: fxa_all_events_table
+        - events_table_view: all_events_table
           base_table: mozdata.firefox_accounts.fxa_all_events
     events:
       type: events_view
@@ -358,6 +358,11 @@ firefox_accounts:
       tables:
         - events_table_view: fxa_content_auth_stdout_events_table
           base_table: mozdata.firefox_accounts.fxa_content_auth_stdout_events
+    all_events_table:
+      type: table_view
+      tables:
+        - table: mozdata.firefox_accounts.fxa_all_events
+          time_partitioning_field: "timestamp"
     fxa_events_table:
       type: table_view
       tables:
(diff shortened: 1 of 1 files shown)


akkomar (approved):

custom-namespaces.yaml
```diff
@@ -338,6 +338,11 @@ firefox_accounts:
     - loines@mozilla.com
   pretty_name: Firefox Accounts
   views:
+    fxa_all_events:
+      type: events_view
+      tables:
+        - events_table_view: fxa_all_events_table
```
sean-rose, line 344: This view also needs to be defined.

custom-namespaces.yaml
```diff
@@ -409,6 +414,11 @@ firefox_accounts:
         - channel: release
           table: mozdata.firefox_accounts.fxa_log_device_command_events
   explores:
+    fxa_all_events:
+      type: events_explore
+      views:
+        base_view: events
```
sean-rose, line 420: This will base it on `fxa_content_auth_oauth_events` (per the `events` view definition above), which isn't what we want.

sean-rose (changes requested):

custom-namespaces.yaml
```diff
@@ -338,6 +338,11 @@ firefox_accounts:
     - loines@mozilla.com
   pretty_name: Firefox Accounts
   views:
+    fxa_all_events:
```
sean-rose, line 341: The existing convention seems to be that the table views have an `fxa_` prefix while the events views (like this one) don't.  So if we want to continue that convention, this would be:
```suggestion
    all_events:
```
(with corresponding change to the explore's `base_view` property)

custom-namespaces.yaml
```diff
@@ -409,6 +414,11 @@ firefox_accounts:
         - channel: release
           table: mozdata.firefox_accounts.fxa_log_device_command_events
   explores:
+    fxa_all_events:
```
sean-rose, line 417: The existing convention for these explore names under "Firefox Acounts" is to not have an `fxa_` prefix, have `event` singular, to have a `_counts` suffix.  So if we want to continue that convention, this would be:
```suggestion
   all_event_counts:
```

kik-kik:
@sean-rose made changes as requested. I hope I did not miss anything this time.

custom-namespaces.yaml
```diff
@@ -353,6 +358,11 @@ firefox_accounts:
       tables:
         - events_table_view: fxa_content_auth_stdout_events_table
           base_table: mozdata.firefox_accounts.fxa_content_auth_stdout_events
+    all_events_table:
+      type: table_view
+      tables:
+        - table: mozdata.firefox_accounts.fxa_all_events
+          time_partitioning_field: "timestamp"
```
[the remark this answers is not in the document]
kik-kik: I might be missing something here, but I do not see a channel field in those views. Here's an example:
image

PR #722, opened 2023-08-02
generate explore for subplat active subscriptions

relud:


custom-namespaces.yaml
```diff
@@ -1183,6 +1183,13 @@ subscription_platform:
       measures:
         count:
           type: count
+    active_subscriptions:
+      type: table_view
+      tables:
+        - table: mozdata.subscription_platform.active_subscriptions
+      measures:
+        count:
+          type: sum
```
sean-rose, line 1192: I believe `sum` measures need a `sql` parameter to specify what to sum.
```suggestion
        count:
          type: sum
          sql: ${count}
```

sean-rose (approved):
r+wc

PR #735, opened 2023-08-25
[AD-78] Update custom-namespaces.yaml

sergiosonline:
Adding consolidated_ad_metrics under new Ads superset

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -519,6 +519,14 @@ ads:
       type: table_view
       tables:
         - table: mozdata.ads.consolidated_ad_metrics
+    consolidated_ad_metrics_daily_pt:
+      type: table_view
+      tables:
+        - table: mozdata.ads.consolidated_ad_metrics_daily_pt
+    consolidated_ad_metrics_hourly:
+      type: table_view
+      tables:
+        - table: mozdata.ads.consolidated_ad_metrics_hourly
 firefox_ios:
   owners:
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -506,6 +506,19 @@ revenue_private:
...
+  pretty_name: Ads
+  owners:
+    - cmorales@mozilla.com
+    - sbetancourt@mozilla.com
+  spoke: looker-spoke-private
+  views:
+    consolidated_ad_metrics:
+      type: table_view
+      tables:
+        - table: mozdata.ads.consolidated_ad_metrics
```
curtismorales, line 521: We should also pull in the Pacific Time table and the hourly table

```suggestion
        - table: mozdata.ads.consolidated_ad_metrics
        - table: mozdata.ads.consolidated_ad_metrics_daily_pt
        - table: mozdata.ads.consolidated_ad_metrics_hourly
```

sergiosonline: @curtismorales : Makes sense. Do we need to add any conditions or filters to separate these three out? Looking at https://github.com/mozilla/lookml-generator/blob/5c771f8c49f39788a597f919f678345904f6abd2/custom-namespaces.yaml#L229

sergiosonline: Yeah, let's do this for now. We can optimize later. Thank you.

curtismorales (approved):
Looks good!

PR #736, opened 2023-08-29
Update to Python 3.10

scholtzan:
Updating to Python 3.10. This is necessary to support generating Looker views/explores for metric-hub using the `metric-config-parser` library. 
Netlify doesn't support Python 3.10, but I think we can disable it. It's been used for showing a preview of the generated `namespace.yaml` file which I added to CircleCI instead.


.circleci/config.yml
```diff
@@ -85,6 +92,7 @@ jobs:
             export GOOGLE_APPLICATION_CREDENTIALS="/tmp/gcp.json"
             echo "$GCLOUD_SERVICE_KEY" > "$GOOGLE_APPLICATION_CREDENTIALS"
             ./bin/generator namespaces
+            cat namespaces.yaml
```
scholtzan, line 95: Show preview of generated file here instead of in netlify.

.circleci/config.yml
```diff
@@ -17,17 +24,17 @@ jobs:
...
+            - python-<< pipeline.parameters.python-version >>-packages-v1-{{ .Branch }}-{{ checksum "requirements.in" }}-
+            - python-<< pipeline.parameters.python-version >>-packages-v1-{{ .Branch }}-
+            - python-<< pipeline.parameters.python-version >>-packages-v1-master-
+            - python-<< pipeline.parameters.python-version >>-packages-v1-
       - &build
         run:
           name: Build
           command: |
-            python3.8 -m venv venv/
+            python3 -m venv venv/
```
kik-kik, line 37: Do we even need to specify python3 binary here? Or is using python instead enough?

Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN groupadd --gid ${USER_ID} ${GROUP_ID} && \
 
 # For grpc https://github.com/grpc/grpc/issues/24556#issuecomment-751797589
 RUN apt-get update -qqy && \
-    apt-get install -qqy python-dev build-essential git curl software-properties-common
+    apt-get install -qqy python3 build-essential git curl software-properties-common
```
kik-kik, line 14: Install python3? This is already a python 3.10 image?

kik-kik (approved):
R+wc

PR #740, opened 2023-08-31
Metric hub integration

scholtzan:
This adds the generation of Looker views and explores for metric definitions defined in metric-hub.
One view is generated for each data source. One explore is generated per namespace that joins these views together on the defined `client_id` and `submission_date` fields.

A branch with the generated Looker artifacts is here: https://github.com/mozilla/looker-hub/tree/ascholtz-metric-hub-1


generator/views/metric_definitions_view.py
```diff
@@ -0,0 +1,207 @@
...
+        # This would result in the selected client_id being set to null (since we always select the client_id)
+        # from the base view.
+        # The following logic selects the first available client_id from any view that is part of the join.
+        joined_client_id_columns = "SAFE_CAST(${TABLE}.client_id AS STRING)"
+        for data_source in MetricsConfigLoader.data_sources_of_namespace(
+            self.namespace
+        ):
+            # for any view that is part of the current query, use the first available client_id
+            joined_client_id_columns += f"""
+                {{% if  metric_definitions_{data_source}._in_query %}}
```
lelilia, line 135:
```suggestion
                {{%- if  metric_definitions_{data_source}._in_query %}}
                , SAFE_CAST(metric_definitions_{data_source}.client_id AS STRING)
                {{%- endif -%}}
```
not sure if the `-` would survive, but currently the formating of the sql is ... not pretty

generator/explores/metric_definitions_explore.py
```diff
@@ -0,0 +1,85 @@
...
+                )
+
+    @staticmethod
+    def from_dict(name: str, defn: dict, views_path: Path) -> MetricDefinitionsExplore:
+        """Get an instance of this explore from a dictionary definition."""
+        return MetricDefinitionsExplore(name, defn["views"], views_path, defn)
+
+    def _to_lookml(
+        self,
+        bq_client: bigquery.Client,
```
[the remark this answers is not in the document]
scholtzan: It's not used in this particular explore, however we are extending the `Explore` class here and we need to keep the same function parameters for the methods we overridde otherwise `mypy` will complain

generator/explores/metric_definitions_explore.py
```diff
@@ -0,0 +1,85 @@
...
+            {
+                "name": joined_view,
+                "view_label": lookml_utils.slug_to_title(joined_view),
+                "relationship": "many_to_many",
+                "type": "full_outer",
+                "sql_on": f"""
+                  SAFE_CAST({self.name}.submission_date AS TIMESTAMP) =
+                  SAFE_CAST({joined_view}.submission_date AS TIMESTAMP) AND
+                  SAFE_CAST({self.name}.client_id AS STRING) =
+                  SAFE_CAST({joined_view}.client_id AS STRING)""",
```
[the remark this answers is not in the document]
scholtzan: Yup, good point. I've made some changes that should fix this (explained in the comment)

generator/explores/explore.py
```diff
@@ -77,11 +77,15 @@ def _to_lookml(
...
-            if not _type.startswith("extended")
-        ]
+        dependent_views = []
+        for _type, views in self.views.items():
+            if _type.startswith("extended"):
+                continue
+            elif _type.startswith("joined"):
+                dependent_views += [view for view in views]
+            else:
+                dependent_views.append(views)
```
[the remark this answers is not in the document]
scholtzan: `joined_views` is a bit of an exception here being a list of views. Usually the `views` specified for an explore are a key value pair with the key and value being strings. So line 87 assumes that `views` is just a string value (which was also what was happening in the removed code bit)

scholtzan:
I made some changes which should help with the join issues and user experience:

If there is a client counts view in the namespace, it will be used as the base view.
The advantage of this is that client counts is guaranteed to have client_ids of all clients that were active on a given day and it exposes some useful fields, like channel, users might want to filter on.

Instead of adding a `date` filter, we can use the `submission_date` field as filter and still inject it into the `derived_table` SQL.

generator/views/metric_definitions_view.py
lelilia: you changed it for the timestamp in line 161 but not for the customer_id :)

generator/explores/metric_definitions_explore.py
```diff
@@ -42,21 +45,22 @@ def from_dict(name: str, defn: dict, views_path: Path) -> MetricDefinitionsExplo
...
     def _to_lookml(
         self,
-        bq_client: bigquery.Client,
-        v1_name: Optional[str],
+        _bq_client: bigquery.Client,
+        _v1_name: Optional[str],
     ) -> List[Dict[str, Any]]:
-        intervals = ["date", "week", "month", "quarter", "year", "raw"]
+        exposed_fields = ["ALL_FIELDS*"]
+        if self.views["base_view"] != "client_counts":
```
lelilia, line 52: Ok I'm a bit confused what is happening here now ...

looking at looker it seems that you are using the `client_counts` view as the base table? which is not part of the whole metrics_definitions_* views

however ... looking at that in looker: [`client_counts`](https://mozilla.cloud.looker.com/projects/looker-hub/files/fenix/views/client_counts.view.lkml) extends [`baseline_clients_daily_table`](https://mozilla.cloud.looker.com/projects/looker-hub/files/fenix/views/baseline_clients_daily_table.view.lkml) which currently has the `client_id` dimension hidden (line 38) 
So currently nowhere in the explore we are able to select a client_id

lelilia: and client_counts has sooo many many fields that I'm not sure I like it for the base fields to be honest ... but that is more of a taste question

scholtzan: Seeing `client_id` isn't super useful in most cases, so having it hidden should be fine.
`client_counts` does have a bunch of dimensions, but a lot of them can be quite useful.

I imagine the explores will be used to answer questions like: What was the total number of active hours (which is a metric coming from metric hub) for all clients on release over a specific time period? What was the average search count (also a metric-hub metric) per country for clients that have FF as default browser?

lelilia: ah ok that does make sense .. I was just primed to look for that now :)

lelilia (approved):
All my concerns are explained or fixed, so I think this looks great \o/

PR #744, opened 2023-09-06
Take 2: Add urlbar events to Looker

rebecca-burwei:
Fixing the name of the explore so it doesn't conflict with the names of existing explores. 

Why I think this will fix it: https://github.com/mozilla/lookml-generator/blob/main/generator/explores/events_explore.py#L40


rebecca-burwei:
@fbertsch I think it's fine to just remove the explore since I have to customize it in looker-spoke-default anyhow.

fbertsch (approved):

PR #763, opened 2023-09-20
Add search_revenue_levers_daily 

skahmann3:


skahmann3:
@alekhyamoz - I'm updating Looker with the latest versions of the search & suggest revenue levers daily tables. It looks like search is currently added to the spoke-private & revenue-private using SQL. What's the best way to maintain this? Should I add to custom-namespaces? Or if I'll have to do the aggregation to get the right summation anyway (specific groupings for reporting), should it be left as is?

alekhyamoz (approved):
LGTM

PR #775, opened 2023-10-07
Add reference dataset to namespaces

curtismorales:
Adds the `reference` dataset and `macroeconomic_indices` view to custom-namespaces

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1062,6 +1062,18 @@ casa:
       type: table_view
       tables:
         - table: mozdata.casa.users
+reference:
+  glean_app: false
+  connection: bigquery-oauth
+  pretty_name: Reference
+  owners:
+    - cmorales@mozilla.com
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -1062,6 +1062,18 @@ casa:
       type: table_view
       tables:
         - table: mozdata.casa.users
+reference:
+  glean_app: false
+  connection: bigquery-oauth
+  pretty_name: Reference
+  owners:
+    - cmorales@mozilla.com
+  spoke: looker-spoke-default
```
lelilia, line 1071: if it's bigquery-oauth connection in has to go into looker-spoke-private

lelilia (changes requested):
the rest looks correct to me

custom-namespaces.yaml
curtismorales: Thanks! I realized I don't the oauth connection

PR #789, opened 2023-10-19
add review checker namespace

betling:
adding review checker namespace to custom-namespaces.yaml with new derived tables.


alekhyamoz (approved):
LGTM

PR #794, opened 2023-10-26
Revert "DENG1781 - Fix urlbar events full release temp view"

alekhyamoz:
Reverts mozilla/lookml-generator#793


rebecca-burwei (commented):
Hi @alekhyamoz - I don't understand the namespace file very well, so perhaps you should get someone else to review. I would think you would want to delete lines 713-721.

PR #798, opened 2023-10-27
android store performance upload

soGaussian:
Adding moz-fx-data-marketing-prod.google_play_store.Store_Performance_country_v1 to looker

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -586,6 +586,12 @@ fenix:
   owners:
     - vzare@mozilla.com
     - ksiegler@mozilla.com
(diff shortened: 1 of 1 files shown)


scholtzan (approved):

fbertsch:
@soGaussian FYI for the `p_Installs_country_v1`, we transformed it in bqetl and we have some norm functions to use: https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/fenix_external/gplay_installs_by_country_v1/query.sql

It might be good to use common approaches for all of these tables.

PR #806, opened 2023-11-07
Generate time dimension groups for datetime metrics in Glean ping views

sean-rose:
This should have been done when we changed the Glean ping views to convert datetime metrics to timestamps (mozilla/bigquery-etl#3453).

Instead, [invalid LookML ended up getting generated](https://github.com/mozilla/looker-hub/commit/2291a2f41fe4fb9d8ea640a83ce84a10bbcf6362) that went unnoticed, until today when Looker started showing `"type": time may only be used in a dimension_group` errors during LookML validation.

### DIFF
--- a/generator/views/glean_ping_view.py
+++ b/generator/views/glean_ping_view.py
@@ -1,5 +1,6 @@
 """Class to describe a Glean Ping View."""
 import logging
+import re
 from collections import Counter
 from textwrap import dedent
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
@@ -293,6 +294,24 @@ class GleanPingView(PingView):
                 },
             ],
         }
+
+        if lookml["type"] == "time":
+            # Remove any _{type} suffix from the dimension group name because each timeframe
+            # will add a _{type} suffix to its individual dimension name.
+            lookml["name"] = re.sub("_(date|time(stamp)?)$", "", looker_name)
+            lookml["timeframes"] = [
+                "raw",
+                "time",
+                "date",
+                "week",
+                "month",
+                "quarter",
+                "year",
+            ]
+            # Dimension groups should not be nested (see issue #82).
+            lookml["label"] = f"{group_label}: {group_item_label}"
+            del lookml["group_label"]
+            del lookml["group_item_label"]
 
         # remove some elements from the definition if we're handling a labeled
         # counter, as an initial join dimension
--- a/generator/views/glean_ping_view.py
+++ b/generator/views/glean_ping_view.py
@@ -309,7 +309,6 @@ class GleanPingView(PingView):
                 "year",
             ]
             # Dimension groups should not be nested (see issue #82).
-            lookml["label"] = f"{group_label}: {group_item_label}"
             del lookml["group_label"]
             del lookml["group_item_label"]
 
--- a/tests/test_glean_ping_view.py
+++ b/tests/test_glean_ping_view.py
@@ -32,6 +32,13 @@ class MockClient:
                                 fields=[SchemaField("fun_url_metric", "STRING")],
                             ),
                             SchemaField(
+                                "datetime",
+                                "RECORD",
+                                fields=[
+                                    SchemaField("fun_datetime_metric", "TIMESTAMP")
+                                ],
+                            ),
+                            SchemaField(
                                 "labeled_counter",
                                 "RECORD",
                                 fields=[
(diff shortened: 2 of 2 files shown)


generator/views/glean_ping_view.py
```diff
@@ -294,6 +295,24 @@ def _make_dimension(
             ],
         }
 
+        if lookml["type"] == "time":
```
lelilia, line 298: Why can't we use the code from the generator here? https://github.com/mozilla/lookml-generator/blob/7a386b47097030f7876361e45ce374314053a93f/generator/views/lookml_utils.py#L42

scholtzan (approved):

generator/views/glean_ping_view.py
sean-rose: Actually that util code does generate dimensions for these datetime metric columns when `super().get_dimensions()` is called, but all metric dimensions from that are currently being excluded in favor of metric dimensions generated here.  One alternate approach would be to not exclude the upstream datetime metric dimensions and not generate datetime metric dimensions here, but there are differences between the two ways of generating dimensions which would be more trouble to work around (labels, hidden/not hidden, links). Overall I felt this was the least complicated way to get the datetime metric dimensions generated correctly that was still consistent with the other metric dimensions.

fbertsch:
Can we add a regression test for this change?

generator/views/glean_ping_view.py
```diff
@@ -294,6 +295,24 @@ def _make_dimension(
...
+            # Remove any _{type} suffix from the dimension group name because each timeframe
+            # will add a _{type} suffix to its individual dimension name.
+            lookml["name"] = re.sub("_(date|time(stamp)?)$", "", looker_name)
```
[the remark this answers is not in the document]
sean-rose: I don't currently see any metric columns that end in `_datetime`, so I don't think that's essential for this PR.

In general I do agree we should also remove `_datetime` suffixes like we do for `_date`, `_time`, and `_timestamp`, but that should be done consistently for all columns, and there are currently some columns with `_datetime` suffixes in `fenix.attributable_clients_v2` and `fenix.firefox_android_clients`, so that would be a potentially breaking change for those which I'd don't want to hassle with in this PR.

If you'd like to create a separate PR for that I'd `r+` it (just make sure to also update the [equivalent code in `lookml_utils.py`](https://github.com/mozilla/lookml-generator/blob/7a386b47097030f7876361e45ce374314053a93f/generator/views/lookml_utils.py#L67)).

generator/views/glean_ping_view.py
```diff
@@ -294,6 +295,24 @@ def _make_dimension(
...
+                "raw",
+                "time",
+                "date",
+                "week",
+                "month",
+                "quarter",
+                "year",
+            ]
+            # Dimension groups should not be nested (see issue #82).
+            lookml["label"] = f"{group_label}: {group_item_label}"
```
fbertsch, line 312: Is this materially different than the previous label? `{group_label} {group_item_label}`

generator/views/glean_ping_view.py
```diff
@@ -294,6 +295,24 @@ def _make_dimension(
...
+            del lookml["group_label"]
+            del lookml["group_item_label"]
```
fbertsch, line 314: rather than adding and deleting, consider removing from the defn above and adding an else clause here

fbertsch (approved):
r+, some optional comments. Thank you for doing this! The long list of the same metric types really annoyed me

generator/views/glean_ping_view.py
sean-rose: I was just following the [precedent from `lookml_utils.py`](https://github.com/mozilla/lookml-generator/blob/7a386b47097030f7876361e45ce374314053a93f/generator/views/lookml_utils.py#L83), but I suppose it's slightly better to be more internally consistent with other metrics here than with that alternate implementation.  I'll remove this.

generator/views/glean_ping_view.py
sean-rose: I actually wrote it that way at first, but ended up deciding this approach was more straightforward/understandable (datetime metrics are only a minor edge case, and the explanatory comment I started writing for the else-group-label logic was annoyingly convoluted compared to the simple "Dimension groups should not be nested" explanation for this logic).

sean-rose:
looker-hub was successfully updated in [commit af7d57c](https://github.com/mozilla/looker-hub/commit/af7d57c9d380305fd90cb0762e627379364dd397), but it turns out link properties aren't supported in dimension groups, so LookML validation is now failing for the same datetime metrics with `Invalid property for "dimension_group": link` errors.  That's being addressed in #807.

PR #808, opened 2023-11-08
COUNT(DISTINCT ${TABLE}.client_id) in client count views

scholtzan:
I've ran into some issues where joining the Client Counts view with other views that had a `client_id` field raised `Column name client_id is ambiguous`

### DIFF
--- a/generator/views/client_counts_view.py
+++ b/generator/views/client_counts_view.py
@@ -63,7 +63,7 @@ class ClientCountsView(View):
             "type": "number",
             "description": "The number of clients, "
             "determined by whether they sent a baseline ping on the day in question.",
-            "sql": "COUNT(DISTINCT client_id)",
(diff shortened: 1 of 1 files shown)


lelilia (approved):

generator/views/client_counts_view.py
```diff
@@ -63,7 +63,7 @@ class ClientCountsView(View):
             "type": "number",
             "description": "The number of clients, "
             "determined by whether they sent a baseline ping on the day in question.",
-            "sql": "COUNT(DISTINCT client_id)",
+            "sql": "COUNT(DISTINCT ${TABLE}.client_id)",
```
fbertsch, line 66: Could also do:
```suggestion
            "sql": "COUNT(DISTINCT ${client_counts.client_id})",
```

fbertsch (approved):

PR #812, opened 2023-11-09
new review checker namespace with mobile

betling:
This creates a new namespace in the custom-namespace yaml file for review checker which includes both mobile and desktop views and explores.


alekhyamoz (approved):
Validated the tables. This looks good to me.
Once the PR is merged and validated, the review_checker_desktop needs to be removed from this file and the corresponding Looker dashboards.

PR #823, opened 2023-12-01
remove summed cards for group by

lelilia:
this might complain in the tests ... at least it does locally ... either flake8 is unhappy or the comparison breaks for test_dashboard_lookml_group_by_dimension ... but it's past 9pm on a friday and I'm out of ideas so ...

### DIFF
--- a/tests/test_operational_monitoring.py
+++ b/tests/test_operational_monitoring.py
@@ -391,3 +391,190 @@ def test_dashboard_lookml(operational_monitoring_dashboard):
     actual = operational_monitoring_dashboard.to_lookml(mock_bq_client)
 
     print_and_test(expected=expected, actual=dedent(actual))
+
+
+@pytest.fixture()
+def operational_monitoring_dashboard_group_by_dimension():
+    return OperationalMonitoringDashboard(
+        "Fission",
+        "fission",
+        "newspaper",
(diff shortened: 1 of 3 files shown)


tests/test_operational_monitoring.py
```diff
@@ -391,3 +391,190 @@ def test_dashboard_lookml(operational_monitoring_dashboard):
...
+    note_text: Mean
+    explore: fission
+    type: looker_line
+    fields: [
+      fission.build_id,
+      fission.branch,
+      fission.point
+    ]
+    pivots: [
+      fission.branch, fission.os
```
lelilia, line 454: here is a whitespace at the end in the result

tests/test_operational_monitoring.py
```diff
@@ -391,3 +391,190 @@ def test_dashboard_lookml(operational_monitoring_dashboard):
...
+    type: "ci-line-chart"
+    fields: [
+      fission.build_id,
+      fission.branch,
+      fission.upper,
+      fission.lower,
+      fission.point
+    ]
+    pivots: [
+      fission.branch, fission.os
```
lelilia, line 492: here is also a whitespace at the end that flake deletes

scholtzan (approved):

PR #841, opened 2023-12-28
Add oauth client ID table and explore

ksiegler1:
Create view and Explore for oauth client ID to name mapping


custom-namespaces.yaml
```diff
@@ -577,6 +581,10 @@ firefox_accounts:
       type: accounts_aggregates_explore
       views:
         base_view: accounts_aggregates
+    accounts_oauth_clients:
+      type: accounts_oauth_clients_explore
```
akkomar, line 585:
```suggestion
      type: table_explore
```

akkomar (approved):

PR #854, opened 2024-01-17
Fix handling event stream tables

scholtzan:
The newly added events stream tables have a string field `event` and an `event_timestamp` field. `event_timestamp` should be handled similar to submission_timestamp.

Currently lookml-generator is failing with `Error: duplicate dimension 'event' for table 'mozdata.org_mozilla_fenix.events_stream'`


kwindau (approved):
lgtm

PR #865, opened 2024-02-05
Generate per-datasource explores for metric definitions

scholtzan:
This PR is making some changes to the `metric_definitions_` views and explores based on the feedback I got from DS and some DEs.
This is still very much a prototype, and something to be tested out by some DS


lelilia (approved):
lgtm
not that I fully understand it though :)

PR #875, opened 2024-02-13
DENG-2611 - Fix Looker PDTs sourced from `telemetry_derived` failing to update

wwyc:
https://mozilla-hub.atlassian.net/browse/DENG-2611

### DIFF
--- a/generator/views/datagroups.py
+++ b/generator/views/datagroups.py
@@ -15,11 +15,11 @@ from generator.views.lookml_utils import BQViewReferenceMap
 
 DEFAULT_MAX_CACHE_AGE = "24 hours"
 
-# Note: INFORMATION_SCHEMA.PARTITIONS has a row with a `last_modified_time` value even for non-partitioned tables.
 SQL_TRIGGER_TEMPLATE = """
-    SELECT MAX(last_modified_time)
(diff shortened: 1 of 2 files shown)


scholtzan (approved):

whd (commented):
Reviewing this made me realize that Looker doesn't actually have the requisite access to run this (since it's per-project metadata, not per-table). That's not a policy issue (e.g. redash does have this access) so I will file a PR to grant the access needed to make this PR work, and update this ticket when that has landed.

generator/views/datagroups.py
```diff
@@ -15,11 +15,11 @@
 
 DEFAULT_MAX_CACHE_AGE = "24 hours"
 
-# Note: INFORMATION_SCHEMA.PARTITIONS has a row with a `last_modified_time` value even for non-partitioned tables.
```
ANich, line 18: I suspect yes, but does this also apply for `INFORMATION_SCHEMA.TABLE_STORAGE`? We probably still want to create datagroups for non-partitioned tables.

ANich (approved):
LGTM, left one question

generator/views/datagroups.py
ANich: That makes sense to me 👍 no need to re-add

whd:
Looker should now have the relevant access.

PR #891, opened 2024-02-23
Update owners

hwine:


custom-namespaces.yaml
```diff
@@ -660,9 +660,7 @@ ads:
 firefox_ios:
   pretty_name: Firefox iOS
   owners:
-    - athomas@mozilla.com
     - darcese@mozilla.com
-    - kik@mozilla.com
```
kik-kik, line 665: Could you provide some context on why I'm being removed from this list?

hwine: Update pushed -- I'll dm you via slack

kik-kik (approved):

PR #895, opened 2024-02-28
[DENG-1908] Auto-import existing explores for new namespaces

BenWu:
https://mozilla-hub.atlassian.net/browse/DENG-1908

This uncomments the explore import for new namespaces added to looker-spoke if explores exist, e.g. the generated ones for glean apps.  @akkomar Is there value in doing this for existing custom namespaces that have explores added?  I think it's reasonable to expect someone to update the spoke repo along with adding explores here.

No new test cases because the existing ones already cover both cases.

### DIFF
--- a/generator/spoke.py
+++ b/generator/spoke.py
@@ -81,11 +81,14 @@ def generate_model(
         "label": namespace_defn["pretty_name"],
     }
 
+    # automatically import generated explores for new glean apps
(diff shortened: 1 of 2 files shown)


akkomar (approved):
> I think it's reasonable to expect someone to update the spoke repo along with adding explores here.

I agree.

PR #907, opened 2024-03-05
Remove `urlbar_clients_daily` from Looker

teonbrooks:
The underlying table has been marked for deprecate via https://github.com/mozilla/bigquery-etl/pull/5046 so we should update Looker to remove the table from the Explore

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -850,17 +850,6 @@ awesome_bar:
   glean_app: false
   owners:
     - ascholtz@mozilla.com
-  pretty_name: Awesome Bar
-  views:
-    urlbar_clients_daily:
-      type: ping_view
-      tables:
-        - table: mozdata.telemetry.urlbar_clients_daily
-  explores:
-    urlbar_clients_daily:
-      type: ping_explore
-      views:
-        base_view: urlbar_clients_daily
 experimentation:
   glean_app: false
   owners:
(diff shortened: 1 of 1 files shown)


alekhyamoz:
This looks good. But the downstream dependencies need to be handled before removing it from the namespace

teonbrooks:
for this View, the only two dependencies it has are two dashboards, one is marked for `Permanently Delete After Nov 30 2023 If Still In Archive Folder` and the second is marked `Not Maintained` and hasn't been changed in a year

teonbrooks:
downstream dependencies have now been taken care of. this is ready to final review and merge

scholtzan (approved):

sean-rose:
The associated namespace/model needs to be removed from [looker-spoke-default](https://github.com/mozilla/looker-spoke-default) as well. This change is causing a LookML validation error due to an include statement in the model no longer matching any files.

sean-rose:
> The associated namespace/model needs to be removed from [looker-spoke-default](https://github.com/mozilla/looker-spoke-default) as well. This change is causing a LookML validation error due to [this include statement](https://github.com/mozilla/looker-spoke-default/blob/9465733fe3ceae6a0292a9068c1771b7ef92caee/awesome_bar/awesome_bar.model.lkml#L4) in the model no longer matching any files.

This has been resolved by https://github.com/mozilla/looker-spoke-default/pull/808. Thanks @scholtzan!

PR #914, opened 2024-03-08
bug 1884384: Add tiktokreporter to disallowlist

edugfilho:


namespaces-disallowlist.yaml
```diff
@@ -11,3 +11,4 @@
 - pioneer_*
 - rally_*
 - reference_browser
+- org_mozilla_tiktokreporter
```
akkomar, line 14:
```suggestion
- tiktokreporter_*
- org_mozilla_ios_tiktok_reporter*
- org_mozilla_tiktokreporter
```
These two additional lines might be needed, but we can try without them.

akkomar (approved):

PR #923, opened 2024-03-20
Support minimal Glean pings

scholtzan:
fixes: https://mozilla.slack.com/archives/C01E8GDG80N/p1710886687720089


quiiver (approved):
i guess this worked even when client_id was not set because it still existed in the schema. lgtm thanks!

PR #927, opened 2024-03-25
Auto-remove expired opmon projects

scholtzan:
This will automatically delete Looker artifacts for opmon projects 90 days after they ended. Right now this would delete views, explores, dashboards for 127 projects.


generator/operational_monitoring_utils.py
```diff
@@ -76,6 +76,9 @@ def get_projects(
             f"""
                 SELECT *
                 FROM `{project_table}`
+                WHERE
+                    DATE_ADD(end_date, INTERVAL 90 DAY) > CURRENT_DATE() OR
+                    end_date IS NULL
```
edugfilho, line 81: Sorry if I missed a warning, but this is going to remove a project I want to keep: [All Browsers DAU](https://mozilla.cloud.looker.com/dashboards/operational_monitoring::all_browsers_dau?Date=&Percentile=99). I wonder if I'm the only case, or should I update it with an end date far in the future?

scholtzan: If you don't set an end date then the project (and Looker dashboards) will be around forever. Yours doesn't have an end date set, so nothing will happen to it

edugfilho: my bad, this is logic to keep 😵

edugfilho (approved):

PR #930, opened 2024-03-26
Update source for Growth Accounting explore

ksiegler1:
Update source of Growth Account explore for Firefox Accounts to use Glean data source, with added service dimension for retention by service.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -531,9 +531,9 @@ firefox_accounts:
           table: mozdata.firefox_accounts.events_daily
     growth_accounting:
       type: growth_accounting_view
-      identifier_field: user_id
+      identifier_field: user_id_sha256
       tables:
-        - table: mozdata.firefox_accounts.fxa_users_last_seen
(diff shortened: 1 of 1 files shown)


kik-kik (approved):

kik-kik (commented):
Before merging we should make sure that there are no downstream resources in looker-spoke-default referencing any fields that no longer exist.

kik-kik (commented):
A quick question, I just had a look at the new table and seeing some weird values under the `service` field:
```
SELECT DISTINCT service
FROM `moz-fx-data-shared-prod.accounts_backend_derived.users_services_last_seen_v1`
WHERE submission_date < "2024-03-27"
```

Example: `749818d3f2e7857f` is this intended (I will share the complete list via Slack)?

kik-kik (changes requested):
Changing to "Request changes" until we can confirm the table is correct.

kik-kik (approved):
Based on conversation with Kimmy.

PR #934, opened 2024-03-28
DENG-2570 - update old GA3 looker views to point to new GA4 looker views

kwindau:


custom-namespaces.yaml
```diff
@@ -1348,23 +1348,23 @@ websites:
     moz_org_page_metrics:
       type: table_view
       tables:
-        - table: moz-fx-data-marketing-prod.ga_derived.www_site_page_metrics_v1
+        - table: moz-fx-data-marketing-prod.ga.www_site_page_metrics
```
kik-kik, line 1351: Should those views be visible and accessible via mozdata project instead?

kwindau: I think because this is in the marketing project, the views are in the marketing project also, not in mozdata

custom-namespaces.yaml
```diff
@@ -435,7 +435,7 @@ mozilla_vpn:
       type: table_view
       tables:
         - channel: release
-          table: mozdata.mozilla_vpn.site_metrics_summary
+          table: mozdata.mozilla_vpn.site_metrics_summary_v2
```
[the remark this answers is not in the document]
kwindau: yes that is correct normally, but this one is a bit of a unique case.  Because we are trying to switch over from GA3 to GA4 but also leave the ability to quickly switch back if needed, we didn't update the view itself to point to the new table, but rather are just switching out the looker view reference.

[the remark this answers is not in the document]
kik-kik: Wouldn't it be easier to make this change inside the view then?

custom-namespaces.yaml
kik-kik: Let me rephrase the question, apart for the reason of how the ingestion of this data is set up. Is there a specific reason why this data is accessible via the marketing project vs out standardised way via moz-prod and mozdata projects?

kwindau: I think mainly because it is marketing data, and that is where the prior data lived also, so for sake of keeping things findable, we kept the GA4 data in the same spot as the GA3 data

PR #936, opened 2024-03-28
Remove Looker artifacts for namespaces that haven't been used in the last 360 days

scholtzan:
It doesn't seem like these have been used or aren't used anymore, so we can remove the Looker content to speed up things a little

### DIFF
--- a/namespaces-disallowlist.yaml
+++ b/namespaces-disallowlist.yaml
@@ -16,3 +16,9 @@
 - org_mozilla_tiktokreporter
 - moso_*
 - regrets_reporter
+- mozphab
+- mozregression
+- pine
+- treeherder
+- sync
+- mozillavpn_backend_cirrus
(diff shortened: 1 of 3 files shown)


kwindau (approved):

namespaces-disallowlist.yaml
```diff
@@ -16,3 +16,9 @@
 - org_mozilla_tiktokreporter
 - moso_*
 - regrets_reporter
+- mozphab
+- mozregression
+- pine
+- treeherder
+- sync
+- mozillavpn_backend_cirrus
```
sean-rose, line 24: `mozillavpn_backend_cirrus` was only recently added ([bug 1867770](https://bugzilla.mozilla.org/show_bug.cgi?id=1867770)) and just hasn't started being used yet.

@brizental are there still plans to use Cirrus with VPN?

brizental: Nope, that whole effort was frozen after the layoffs. We might pick it up agai nin the future, but that is unclear at the moment. If this can be reverted easily in the future, it's fine by me.

PR #952, opened 2024-04-15
Remove Cirrus apps

scholtzan:
As discussed with the Nimbus data leads, we don't need this data in Looker

### DIFF
--- a/namespaces-disallowlist.yaml
+++ b/namespaces-disallowlist.yaml
@@ -24,3 +24,4 @@
 - mozillavpn_backend_cirrus
 - glean_dictionary
 - mach
+- "*_cirrus"


namespaces-disallowlist.yaml
```diff
@@ -24,3 +24,4 @@
 - mozillavpn_backend_cirrus
 - glean_dictionary
 - mach
+- "*_cirrus"
```
scholtzan, line 27: `""` needed here otherwise yaml thinks this is a reference

mikewilli (approved):
LGTM

Does it make sense to (also, separately) comment out lines for the cirrus apps in looker-spoke-default ([example](https://github.com/mozilla/looker-spoke-default/blob/main/accounts_cirrus/accounts_cirrus.model.lkml))? Or at least add a note to that effect in the runbook so they aren't enabled in the first place when adding new Glean apps?

scholtzan:
The spoke-default files will get removed automatically on the next lookml-generator run

PR #960, opened 2024-04-25
DENG-3540 Add `event_id` as a primary key for event view

quiiver:
UPDATE:
This has been updated to use the `event_id` column that was recently added to `events_unnested`. This gets around the uniqueness problems listed below

---

In the generated events explore for glean apps it will incorrectly count events if the explore is using unnested joins for either event extras or experiments. When the those joins are present looker counts events by the tables primary key. The primary key that's defined for those tables is `document_id`. Unfortunately, that's not a good primary key as there can be multiple events per ping. 

In this PR i'm creating a synthetic primary key by combining document_id and event_timestamp. That is also not a perfect event_id but it is much better. For example on April 1st, counting distinct doc_ids in fx desktop return a count of 213 million; counting distinct doc_id + event_ts returned a count of 2.878 billion. That's a difference of 2.6 billion. In practice the difference won't be so drastic but that illustrates the problem.

As i mentioned, using just timestamp and document id isn't perfect. If we include event_name we count an additional 5 million for the day. However, there is a cost. Adding that additional Identifier over doubles the bytes scanned. Going from 34gb to 88gb.

Test queries here:
```sql
with counts as (
    SELECT
        COUNT(DISTINCT glean_events.document_id ) AS doc_ids,
        COUNT(DISTINCT concat(glean_events.document_id, glean_events.event_timestamp) ) as event_id_ts,
        COUNT(DISTINCT concat(glean_events.document_id, glean_events.event_name, glean_events.event_timestamp) ) as event_id_ts_name, 
        COUNT(DISTINCT concat(glean_events.document_id, glean_events.event_category, glean_events.event_name, glean_events.event_timestamp) ) as event_id_cat_name_ts 
    FROM `mozdata.firefox_desktop.events_unnested`  AS glean_events
    LEFT JOIN UNNEST(glean_events.event_extra) AS events_unnested_table__event_extra
    WHERE DATE(glean_events.submission_timestamp ) = '2024-04-01' 
)

select 
    *,
    FORMAT("%'d", event_id_ts-doc_ids) as doc_id_diff,
    FORMAT("%'d", counts.event_id_ts_name - counts.event_id_ts) as ts_name_diff,
    FORMAT("%'d", counts.event_id_cat_name_ts - counts.event_id_ts_name) as cat_name_ts_diff,
from counts
```


generator/views/events_view.py
```diff
@@ -104,3 +102,16 @@ def get_measures(self, dimensions) -> List[Dict[str, str]]:
...
+    def generate_event_id(self, dimensions: list[dict]) -> Optional[Dict[str, str]]:
+        """Generate the event_id dimension to be used as a primary key for a one to many join."""
+        document_id_field = self.get_document_id(dimensions, "events")
+        event_timestamp = self.select_dimension("event_timestamp", dimensions, "events")
+        event_timestamp_field = event_timestamp["name"] if event_timestamp else None
+        if document_id_field and event_timestamp_field:
+            return {
+                "name": "event_id",
+                "primary_key": "yes",
+                "sql": f"CONCAT(${{{document_id_field}}}, ${{{event_timestamp_field}}})",
```
sean-rose, line 115: - Rather than `event_timestamp` (which can be the same for multiple events from the same document/ping) I think it would be preferable to have a guaranteed unique ID per event, which I think we can easily achieve using the `UNNEST ... WITH OFFSET AS ...` syntax in `events_unnested`.  I'll file a PR to add a column based on such an offset to `events_unnested` shortly.
- I'd suggest adding a `-` separator between the two values for readability in case us humans want to look at the values.

generator/views/view.py
```diff
@@ -104,26 +104,31 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
...
+        selected = [d for d in dimensions if d["name"] in dimension_names]
+        if not selected:
             # Some pings purposely disinclude client_ids, e.g. firefox installer
             return None
-        if len(document_id_fields) > 1:
-            raise ClickException(f"Duplicate document_id dimension in {table!r}")
-        return document_id_fields[0]
+        if len(selected) > 1:
+            raise ClickException(f"Duplicate {dimension_names} dimension in {table!r}")
+        return selected[0]
```
sean-rose, line 134: If you're looking for multiple dimension names, saying there's a duplicate if more than one of those dimensions is present doesn't seem accurate. I also think it'd be good to allow for callers to pass a prioritized list of dimensions, checking for those in the order the caller specified (implying the use of a list or other sequence rather than an unordered set), and returning the first one found.

So perhaps something like this instead:
```python
    def select_dimension(
        self,
        dimension_names: str | List[str],
        dimensions: List[dict],
        table: str,
    ) -> Optional[dict[str, str]]:
        """Return the first field that matches dimension name."""
        if isinstance(dimension_names, str):
            dimension_names = [dimension_names]
        for dimension_name in dimension_names:
            selected = [d for d in dimensions if d["name"] == dimension_name]
            if selected:
                if len(selected) > 1:
                    raise ClickException(f"Duplicate {dimension_name} dimension in {table!r}")
                return selected[0]
        return None
```

generator/views/view.py
```diff
@@ -104,26 +104,31 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
...
-            return None
-        if len(client_id_fields) > 1:
-            raise ClickException(f"Duplicate client_id dimension in {table!r}")
-        return client_id_fields[0]
+        client_id_fields = self.select_dimension(
+            {"client_id", "client_info__client_id", "context_id"},
+            dimensions,
+            table,
+        )
+        return client_id_fields["name"] if client_id_fields else None
```
sean-rose, line 112: I think this comment is more appropriate here:
```suggestion
        # Some pings purposely disinclude client_ids, e.g. firefox installer
        return client_id_fields["name"] if client_id_fields else None
```

generator/views/events_view.py
quiiver: yes offset is a great idea. thanks!

sean-rose: #5446

chelseybeck:
For creating a unique primary key, I've seen it work well like this before:

TO_HEX(SHA256(to_json_STRING(STRUCT({col1, col2, col3})))) AS unique_id

Then, it's based on whichever columns you select (which are likely to make differences) and they get hashed together

generator/views/view.py
quiiver: Ok so I started to refactor this and it seems like we want the old behavior which is "Select the first item in the set, throw if more than one item in the set is selected"

sean-rose: OK, I now see how you're preserving the previous behavior of `get_client_id()`.

generator/views/view.py
```diff
@@ -104,26 +104,39 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
...
+        """
+        if isinstance(dimension_names, str):
+            dimension_names = {dimension_names}
+        selected = [d for d in dimensions if d["name"] in dimension_names]
+        if selected:
+            # there should only be one dimension selected from the set
+            # if there are multiple options in the dimention_names set.
+            if len(dimension_names) > 1 and len(selected) > 1:
+                raise ClickException(
+                    f"Duplicate {dimension_names} dimension in {table!r}"
```
sean-rose, line 139:
```suggestion
                    f"Duplicate {'/'.join(dimension_names)} dimensions in {table!r}"
```

generator/views/events_view.py
```diff
@@ -76,11 +76,9 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
 
         # set document_id as primary key if it exists in the underlying table
         # this will allow one_to_many joins
-        document_id_field = self.get_document_id(dimensions, "events")
-        if document_id_field is not None:
-            view_defn["dimensions"] = [
-                {"name": document_id_field, "primary_key": "yes"}
-            ]
```
scholtzan, line 83: Would this remove `document_id` from the views/explores? Should we still keep the field around in case any existing looks or dashboards are referencing it?

quiiver: I'll confirm but I don't think it removes it. It should still be present in the underlying `_table` view that this extends.

quiiver: It's defined here: https://github.com/mozilla/looker-hub/blob/main/firefox_desktop/views/events_stream_table.view.lkml#L116-L119

scholtzan (approved):

PR #981, opened 2024-05-22
Fix date filters for metrics definitions views

scholtzan:
Fixes https://mozilla.slack.com/archives/C01J4BPCYGM/p1716394737699449

### DIFF
--- a/generator/views/metric_definitions_view.py
+++ b/generator/views/metric_definitions_view.py
@@ -169,11 +169,11 @@ class MetricDefinitionsView(View):
                     BETWEEN
                     COALESCE(
                         SAFE_CAST(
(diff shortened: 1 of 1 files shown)


generator/views/metric_definitions_view.py
```diff
@@ -169,11 +169,11 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
                     BETWEEN
                     COALESCE(
                         SAFE_CAST(
-                            {{% date_start {data_source.submission_date_column or "submission_date"} %}} AS DATE
```
scholtzan, line 172: A `submission` dimension is always created pointing to `submission_date_column`, so we don't need to do the mapping here

kwindau (approved):

PR #992, opened 2024-06-07
Add --only parameter to generate lookml only for select namespaces

scholtzan:
For testing some changes locally it would be convenient if it's possible to select specific namespaces that should be generated

### DIFF
--- a/generator/lookml.py
+++ b/generator/lookml.py
@@ -100,7 +100,7 @@ def _glean_apps_to_v1_map(glean_apps):
     return {d["name"]: d["v1_name"] for d in glean_apps}
 
 
-def _lookml(namespaces, glean_apps, target_dir):
+def _lookml(namespaces, glean_apps, target_dir, namespace_filter=[]):
     client = bigquery.Client()
 
     namespaces_content = namespaces.read()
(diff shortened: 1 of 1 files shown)


generator/lookml.py
```diff
@@ -115,37 +115,40 @@ def _lookml(namespaces, glean_apps, target_dir):
...
-
-        logging.info("  Generating dashboards")
-        dashboard_dir = target / namespace / "dashboards"
-        dashboard_dir.mkdir(parents=True, exist_ok=True)
-        dashboards = lookml_objects.get("dashboards", {})
-        for dashboard_path in _generate_dashboards(
-            client, dashboard_dir, namespace, dashboards
-        ):
-            logging.info(f"    ...Generating {dashboard_path}")
+        if len(namespace_filter) == 0 or namespace in namespace_filter:
```
BenWu, line 118: Should check for `None`?
```suggestion
        if namespace_filter is None or namespace in namespace_filter:
```
or the option can default to `[]`

BenWu (commented):
Unit test needs to be updated with the new argument to `_lookml`

generator/lookml.py
scholtzan: yeah, defaulting to `[]` is probably better

BenWu (approved):

PR #994, opened 2024-06-11
Parallel lookml generation

scholtzan:
This makes some changes to the lookml generation to generate views, explores and dashboards in parallel. Not all of the generation can run in parallel since, for example, the explore generation depends on generated view files.

In the tests that I ran locally, this speeds up lookml generation by 2x

### DIFF
--- a/generator/lookml.py
+++ b/generator/lookml.py
@@ -1,6 +1,8 @@
 """Generate lookml from namespaces."""
 
 import logging
+from functools import partial
+from multiprocessing.pool import ThreadPool
 from pathlib import Path
 from typing import Dict, Iterable, Optional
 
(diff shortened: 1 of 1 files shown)


Marlene-M-Hirose (approved):

generator/lookml.py
```diff
@@ -126,29 +130,51 @@ def _lookml(namespaces, glean_apps, target_dir, namespace_filter=[]):
...
             v1_name: Optional[str] = v1_mapping.get(namespace)
-            for view_path in _generate_views(client, view_dir, views, v1_name):
-                logging.info(f"    ...Generating {view_path}")
+            for view in views:
+                views_with_v1_name.append(
+                    partial(_generate_view, client, view_dir, view, v1_name)
+                )
+
+    with ThreadPool(parallelism) as pool:
+        pool.map(lambda x: x(), views_with_v1_name, chunksize=1)
```
sean-rose, line 139: It'd be nice to have a comment here explaining why `pool.map()` with `chunksize=1` was called for, compared to the other cases which just use `pool.starmap()`.

scholtzan: In this case `starmap` would actually be slightly more readable

PR #998, opened 2024-06-21
Update Google Search Console views to use BigQuery views in `mozdata`

sean-rose:
Because the BigQuery views are being moved (mozilla/bigquery-etl#5822).


kwindau (approved):
LGTM

PR #1003, opened 2024-07-04
Use dry run function for generation

scholtzan:
Depends on https://github.com/mozilla-services/cloudops-infra/pull/5769

This adds a `use_cloud_function` option for lookml-generator to use the dryrun Cloud Function for the LookML generation. This solves multiple problems:
* Generating LookML for restricted tables. When running generation locally with user credentials the generation currently fails as access is restricted to tables such as `ads`
* Generating LookML in CI. This is desirable to have to generate change diffs.

I generated a diff here: https://github.com/mozilla/looker-hub/compare/test-lookml-generation-dryrun?expand=1
The main changes are in the descriptions to use `.` instead of `:` for table IDs. `user_journey` artifacts show up as deleted which is [expected since none are defined in `custom-namespaces.yaml`](https://github.com/mozilla/lookml-generator/blob/8a5bee98eeeef7ded36f739cb4b06677247789e7/custom-namespaces.yaml#L1584-L1588)

The dry run function cannot generate `operational_monitoring` namespace information as this requires querying information from some tables in BigQuery. When we start using this in CI to generate diff, this namespace will need to be skipped using the `--ignore` option.

### DIFF
--- a/generator/__init__.py
+++ b/generator/__init__.py
@@ -6,17 +6,36 @@
 
 __docformat__ = "restructuredtext"
 
+import sys
 import warnings
 
 import click
+from google.auth.exceptions import DefaultCredentialsError
+from google.cloud import bigquery
 
 from .lookml import lookml
 from .namespaces import namespaces
 from .spoke import update_spoke
 
 
+def is_authenticated():
+    """Check if the user is authenticated to GCP."""
+    try:
+        bigquery.Client()
+    except DefaultCredentialsError:
+        return False
+    return True
+
+
 def cli(prog_name=None):
     """Generate and run CLI."""
+    if not is_authenticated():
+        print(
+            "Authentication to GCP required. Run `gcloud auth login --update-adc` "
+            "and check that the project is set correctly."
+        )
+        sys.exit(1)
+
     commands = {
         "namespaces": namespaces,
         "lookml": lookml,
--- a/generator/dryrun.py
+++ b/generator/dryrun.py
@@ -36,11 +36,12 @@ def id_token():
 class DryRunError(Exception):
     """Exception raised on dry run errors."""
 
-    def __init__(self, message, error, use_cloud_function):
+    def __init__(self, message, error, use_cloud_function, table_id):
         """Initialize DryRunError."""
         super().__init__(message)
         self.error = error
         self.use_cloud_function = use_cloud_function
+        self.table_id = table_id
 
 
 class Errors(Enum):
@@ -160,7 +161,10 @@ class DryRun:
         """Return the query schema by dry running the SQL file."""
         if not self.is_valid():
             raise DryRunError(
-                "Error when dry running SQL", self.get_error(), self.use_cloud_function
+                "Error when dry running SQL",
+                self.get_error(),
+                self.use_cloud_function,
+                self.table,
             )
 
         if (
@@ -176,7 +180,10 @@ class DryRun:
         """Return the schema of the provided table."""
         if not self.is_valid():
             raise DryRunError(
-                "Error when dry running SQL", self.get_error(), self.use_cloud_function
+                "Error when dry running SQL",
+                self.get_error(),
+                self.use_cloud_function,
+                self.table,
             )
 
         if (
@@ -192,7 +199,10 @@ class DryRun:
         """Return table metadata."""
         if not self.is_valid():
             raise DryRunError(
-                "Error when dry running SQL", self.get_error(), self.use_cloud_function
+                "Error when dry running SQL",
+                self.get_error(),
+                self.use_cloud_function,
+                self.table,
             )
 
         if (
@@ -255,6 +265,9 @@ class DryRun:
                 in error_message
             ):
                 return Errors.DATE_FILTER_NEEDED_AND_SYNTAX
-            if "Permission bigquery.tables.get denied on table" in error_message:
+            if (
+                "Permission bigquery.tables.get denied on table" in error_message
+                or "User does not have permission to query table" in error_message
+            ):
                 return Errors.PERMISSION_DENIED
         return None
--- a/generator/dryrun.py
+++ b/generator/dryrun.py
@@ -1,0 +1,218 @@
+from google.cloud import bigquery
+from google.auth.exceptions import DefaultCredentialsError
+import google.auth
+from google.auth.transport.requests import Request as GoogleAuthRequest
+from google.cloud import bigquery
+from google.oauth2.id_token import fetch_id_token
+from urllib.request import Request, urlopen
+import json
+from functools import cached_property
+from enum import Enum
+from typing import Optional
+
+import sys
+
+DRY_RUN_URL = (
+    "https://us-central1-moz-fx-data-shared-prod.cloudfunctions.net/bigquery-etl-dryrun"
+)
+
+
+def is_authenticated():
+    """Check if the user is authenticated to GCP."""
+    try:
+        bigquery.Client(project="")
+    except DefaultCredentialsError:
+        return False
+    return True
+
+
+class Errors(Enum):
+    """DryRun errors that require special handling."""
+
+    READ_ONLY = 1
+    DATE_FILTER_NEEDED = 2
+    DATE_FILTER_NEEDED_AND_SYNTAX = 3
+
+
+class DryRun:
+    """Dry run SQL."""
+
+    def __init__(
+        self,
+        sql=None,
+        use_cloud_function=True,
+        client=None,
+        project=None,
+        dataset=None,
+        table=None,
+        dry_run_url=DRY_RUN_URL,
+    ):
+        self.sql = sql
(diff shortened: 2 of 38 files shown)


generator/lookml.py
```diff
@@ -182,10 +183,22 @@ def _lookml(namespaces, glean_apps, target_dir, namespace_filter=[]):
...
     """Generate lookml from namespaces."""
     if metric_hub_repos:
         MetricsConfigLoader.update_repos(metric_hub_repos)
 
     glean_apps = _get_glean_apps(app_listings_uri)
-    return _lookml(namespaces, glean_apps, target_dir, only)
+
+    dryrun = functools.partial(
+        DryRun, bigquery.Client(), use_cloud_function, id_token()
+    )
```
scholtzan, line 203: Creating a partially applied `DryRun` instance and passing it turns out to be most efficient. 
Creating a `bigquery.Client` instance comes with some overhead, so doing it only once saves quite a bit of time (2x increase when creating a new client instance before each dryrun). Same goes for creating the ID token which is needed to invoke the dryrun cloud function (2x increase when creating the token before each dryrun)

generator/namespaces.py
```diff
@@ -350,23 +364,33 @@ def namespaces(
...
+            del custom_namespaces[ignored_namespace]
 
         # generating operational monitoring namespace, if available
         if "operational_monitoring" in custom_namespaces:
+            if use_cloud_function:
+                raise Exception(
+                    "Cannot generate OpMon using dry run Cloud Function"
+                    + "`export USE_CLOUD_FUNCTION=False` to use user GCP credentials"
+                )
+
```
scholtzan, line 393: operational monitoring is actually querying some tables to get information that is needed for generating artifacts. With the dry run function these artifacts cannot be generated. 
This is something to be addressed in a future PR.

generator/utils.py
```diff
@@ -0,0 +1,20 @@
...
+    file = path.name
+    artifact_type = path.parent.name
+    namespace = path.parent.parent.name
+    print(f"{LOOKER_HUB_URL}/{namespace}/{artifact_type}/{file}")
+    with urllib.request.urlopen(
+        f"{LOOKER_HUB_URL}/{namespace}/{artifact_type}/{file}"
+    ) as response:
+        lookml = response.read().decode(response.headers.get_content_charset())
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(lookml)
```
scholtzan, line 20: For cases where the dryrun function runs into a permission error, it will check whether the file exists in looker-hub and simply copy it. So this would be a noop. 
This is potentially not ideal in cases where the logic of these file does and should change as it would appear as if the changes didn't take effect. We should probably highlight affected tables to users if/when we post some kind of diff in the PR.

generator/__init__.py
```diff
@@ -6,17 +6,36 @@
...
 from .spoke import update_spoke
 
 
+def is_authenticated():
+    """Check if the user is authenticated to GCP."""
+    try:
+        bigquery.Client()
+    except DefaultCredentialsError:
+        return False
+    return True
```
sean-rose, line 27: I see this is using the [same `is_authenticated()` logic as BigQuery ETL](https://github.com/mozilla/bigquery-etl/blob/f86a8d7ac197ac32d65badc7228966877199c829/bigquery_etl/cli/utils.py#L42-L48), which is fine for now, but it has the same drawback where after Google application default credentials have been set at least once this will always return true, and at runtime you'll get a `google.auth.exceptions.RefreshError: Reauthentication is needed` error.

So if/when we eventually address that drawback for BigQuery ETL we should remember to do so here as well.

generator/namespaces.py
```diff
@@ -350,23 +364,33 @@ def namespaces(
...
+                "views": views_as_dict,
+                "explores": explores,
+                "glean_app": True,
+            }
 
     if custom_namespaces is not None:
         custom_namespaces = yaml.safe_load(custom_namespaces.read()) or {}
+        # remove namespaces that should be ignored
+        for ignored_namespace in ignore:
+            del custom_namespaces[ignored_namespace]
```
sean-rose, line 384: This needs another check, otherwise if someone passes a namespace to ignore that isn't a custom namespace this would raise a `KeyError`.
```suggestion
            if ignored_namespace in custom_namespaces:
                del custom_namespaces[ignored_namespace]
```

generator/views/datagroups.py
```diff
@@ -153,14 +186,27 @@ def generate_datagroups(
...
+                datagroup := _generate_view_datagroup(view, dataset_view_map, dryrun)
+            ) is not None:
+                datagroups.add(datagroup)
+        except DryRunError as e:
+            if e.error == Errors.PERMISSION_DENIED and e.use_cloud_function:
+                path = (
+                    datagroups_folder_path / f"{e.table_id}_last_updated.datagroup.lkml"
+                )
+                print(
+                    f"Permission error dry running: {path}. Copy existing file from looker-hub."
```
sean-rose, line 202: This message doesn't quite make sense to me.  Maybe something like this?
```suggestion
                    f"Permission error dry running {view.name}. Copy existing {path} file from looker-hub."
```

generator/dryrun.py
```diff
@@ -0,0 +1,273 @@
...
+            )
+
+        if (
+            self.dry_run_result
+            and self.dry_run_result["valid"]
+            and "schema" in self.dry_run_result
+        ):
+            return self.dry_run_result["schema"]["fields"]
+
+        return {}
```
sean-rose, line 177: Shouldn't this be an empty list?
```suggestion
        return []
```

generator/dryrun.py
```diff
@@ -0,0 +1,273 @@
...
+            )
+
+        if (
+            self.dry_run_result
+            and self.dry_run_result["valid"]
+            and "tableMetadata" in self.dry_run_result
+        ):
+            return self.dry_run_result["tableMetadata"]["schema"]["fields"]
+
+        return {}
```
sean-rose, line 196: Shouldn't this be an empty list?
```suggestion
        return []
```

generator/views/metric_definitions_view.py
```diff
@@ -283,7 +285,10 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
         return {"views": [view_defn]}
 
     def get_dimensions(
-        self, _bq_client=None, _table=None, _v1_name: Optional[str] = None
+        self,
+        _table=None,
+        _v1_name: Optional[str] = None,
+        dryrun=False,
```
sean-rose, line 291: To match the convention of the other unused arguments should `dryrun` be prefixed with an underscore and have its default value be `None`?
```suggestion
        _dryrun=None,
```

generator/views/metric_definitions_view.py
```diff
@@ -253,7 +255,7 @@ def to_lookml(self, bq_client, v1_name: Optional[str]) -> Dict[str, Any]:
             """
         }
 
-        view_defn["dimensions"] = self.get_dimensions()
+        view_defn["dimensions"] = self.get_dimensions(dryrun=dryrun)
```
sean-rose, line 258: I don't think `dryrun` should be passed here since this `get_dimensions()` implementation doesn't use it.
```suggestion
        view_defn["dimensions"] = self.get_dimensions()
```

generator/lookml.py
sean-rose: Should we do something similar in BigQuery ETL then?

scholtzan: yeah, I'd be worth investigating whether we can make some improvements there

generator/namespaces.py
```diff
@@ -350,23 +364,33 @@ def namespaces(
...
 
         # generating operational monitoring namespace, if available
         if "operational_monitoring" in custom_namespaces:
+            if use_cloud_function:
+                raise Exception(
+                    "Cannot generate OpMon using dry run Cloud Function"
+                    + "`export USE_CLOUD_FUNCTION=False` to use user GCP credentials"
```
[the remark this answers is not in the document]
scholtzan: forgot to remove this. initially I had an implementation that was using env variables

generator/lookml.py
```diff
@@ -26,24 +30,39 @@
...
+        except DryRunError as e:
+            if e.error == Errors.PERMISSION_DENIED and e.use_cloud_function:
+                print(
+                    f"Permission error dry running {view.name}. Copy existing {path} file from looker-hub."
+                )
+                try:
+                    get_file_from_looker_hub(path)
+                except Exception as ex:
+                    print(f"Skip generating view for {path}: {ex}")
+                yield path
```
sean-rose, line 60:
```suggestion
                try:
                    get_file_from_looker_hub(path)
                    yield path
                except Exception as ex:
                    print(f"Skip generating view for {path}: {ex}")
```

sean-rose (approved):
r+wc

PR #1007, opened 2024-07-15
Rename `event` to `full_event_name` in `events_stream` views

akkomar:


generator/views/lookml_utils.py
```diff
@@ -117,6 +117,16 @@ def _generate_dimensions(client: bigquery.Client, table: str) -> List[Dict[str,
...
     for dimension in _generate_dimensions_helper(client.get_table(table).schema):
+        # Rename `event` field in `events_stream` to `full_event_name` to expose it in Looker views.
+        # This is required since `event_timestamp` is renamed to `event` in `_get_dimension` and takes precedence below.
+        if (
+            table.endswith(".events_stream")
+            and dimension["name"] == "event"
+            and dimension["type"] == "string"
+        ):
+            # `event` field contains concatenated event category and name
+            dimension["name"] = "full_event_name"
```
BenWu, line 128: Can the `and not (name == "event" and dimension["type"] == "time")` line below be removed since this is being renamed?  Looks like that's just for events stream https://github.com/mozilla/lookml-generator/pull/854 and might be useful to catch other similar issues.  Not critical though and maybe it's better to leave it as-is

BenWu (approved):

generator/views/lookml_utils.py
sean-rose, line 128: Why `full_event_name` instead of simply `event_name`?

generator/views/lookml_utils.py
```diff
@@ -117,6 +117,16 @@ def _generate_dimensions(client: bigquery.Client, table: str) -> List[Dict[str,
     """
     dimensions = {}
     for dimension in _generate_dimensions_helper(client.get_table(table).schema):
+        # Rename `event` field in `events_stream` to `full_event_name` to expose it in Looker views.
+        # This is required since `event_timestamp` is renamed to `event` in `_get_dimension` and takes precedence below.
```
sean-rose, line 121: IMO it would be better to change the other logic to not have dimension groups conflict in this way, since they wouldn't actually conflict in the generated LookML, as dimension groups generate dimensions with timeframe suffixes.

akkomar: https://github.com/mozilla/lookml-generator/pull/1010

PR #1010, opened 2024-07-16
Allow same names for dimensions and time dimension groups in generated views

akkomar:
Alternative to https://github.com/mozilla/lookml-generator/pull/1007 that would fix https://github.com/mozilla/bigquery-etl/issues/5903, implements suggestion from https://github.com/mozilla/lookml-generator/pull/1007#discussion_r1678304804.

This approach seems to be minimally invasive and generalized, without hardcoding table names, allowing to unblock https://github.com/mozilla/bigquery-etl/issues/5903.

### DIFF
--- a/generator/views/lookml_utils.py
+++ b/generator/views/lookml_utils.py
@@ -117,25 +117,25 @@ def _generate_dimensions(client: bigquery.Client, table: str) -> List[Dict[str, 
     """
     dimensions = {}
     for dimension in _generate_dimensions_helper(client.get_table(table).schema):
-        name = dimension["name"]
+        name_key = dimension["name"]
 
         # This prevents `time` dimension groups from overwriting other dimensions below
         if dimension.get("type") == "time":
-            name += "_time"
+            name_key += "_time"
 
         # overwrite duplicate "submission", "end", "start" dimension group, thus picking the
         # last value sorted by field name, which is submission_timestamp
         # See also https://github.com/mozilla/lookml-generator/issues/471
         if (
-            name in dimensions
-            and name != "submission"
-            and not name.endswith("end")
-            and not name.endswith("start")
+            name_key in dimensions
+            and dimension["name"] != "submission"
+            and not dimension["name"].endswith("end")
+            and not dimension["name"].endswith("start")
         ):
             raise click.ClickException(
-                f"duplicate dimension {name!r} for table {table!r}"
+                f"duplicate dimension {name_key!r} for table {table!r}"
             )
-        dimensions[name] = dimension
+        dimensions[name_key] = dimension
     return list(dimensions.values())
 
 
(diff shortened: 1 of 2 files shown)


generator/views/lookml_utils.py
```diff
@@ -117,23 +117,25 @@ def _generate_dimensions(client: bigquery.Client, table: str) -> List[Dict[str,
...
         # last value sorted by field name, which is submission_timestamp
         # See also https://github.com/mozilla/lookml-generator/issues/471
         if (
-            name in dimensions
-            and name != "submission"
-            and not name.endswith("end")
-            and not name.endswith("start")
-            and not (name == "event" and dimension["type"] == "time")
-            # workaround for `mozdata.firefox_desktop.desktop_installs`
-            and not (name == "attribution_dltoken" and dimension["type"] == "time")
```
akkomar, line 131: We no longer need these now.

generator/views/lookml_utils.py
```diff
@@ -117,23 +117,25 @@ def _generate_dimensions(client: bigquery.Client, table: str) -> List[Dict[str,
...
+            name_key in dimensions
+            and dimension["name"] != "submission"
+            and not dimension["name"].endswith("end")
+            and not dimension["name"].endswith("start")
```
sean-rose, line 133: I'd be inclined to have these exceptions more explicitly targeted at dimension groups, as the relatively broad `endswith("end")` and `endswith("start")` conditions make me a bit nervous.  Something like
```suggestion
            and name_key != "submission_time"
            and not name_key.endswith("end_time")
            and not name_key.endswith("start_time")
```
or
```suggestion
            and not (
                dimension.get("type") == "time"
                and (
                    dimension["name"] == "submission" 
                    or dimension["name"].endswith("end")
                    or dimension["name"].endswith("start")
                )
            )
```

generator/views/lookml_utils.py
```diff
@@ -145,21 +147,23 @@ def _generate_dimensions_from_query(
...
```
sean-rose, line 163: Same here.

It seems like it'd be good to refactor this repeated logic in `_generate_dimensions` and `_generate_dimensions_from_query` out into a separate function named something like `_generate_dimensions_from_schema`. The only catch is the exception messages for duplicate dimensions are different, and in `_generate_dimensions` it does helpfully include the table in the error message, but for that case you could probably wrap the `_generate_dimensions_from_schema` call in a try/catch and do something like
```python
raise click.ClickException(f"Error generating dimensions for table {table!r}") from e
```

sean-rose (approved):
r+wc

I appreciate you being willing to pivot to this alternate approach after having already coded your original approach.

generator/views/lookml_utils.py
akkomar: Yeah, this part is confusing, narrowing this check down won't hurt.

PR #1017, opened 2024-07-25
Add custom namespace for Firefox-CI Taskcluster

ahal:


scholtzan (approved):
Do you also want to have explores for these views? If so you can have them generated by adding something like this: https://github.com/mozilla/lookml-generator/blob/af9330c33b8df2a4d5bea803b5bdb125275ddb7c/custom-namespaces.yaml#L65

If you are planning to manually create a custom explore or custom views from these generated views then generating the explores might not be necessary. But if you want to expose the table referenced in these views just as they are in Looker, then having these explores generated so you can access the data is good practice.

ahal:
> Do you also want to have explores for these views? If so you can have them generated by adding something like this

Thanks, yes I think I'll add this. For now I just want something to start trying things out. Though I may try adding custom explores later.

ahal:
@scholtzan thanks for the suggestion, could you please merge this for me when you get a chance?

PR #1021, opened 2024-07-29
Parallelize lookml generation

scholtzan:
This PR speeds up the lookml generation by running some of the generation tasks in parallel.

Using the cloud function, generation completes in about 5 minutes. Using service credentials, generation takes about 10 minutes as the Bigquery client needs to be instantiated for every process as it cannot be shared across processes. This is a significant improvement over the 30+ minutes the generation currently takes. 

One drawback is that if an error happens during generation, it might take longer for the generator to exit as it will wait for all processes to complete

I ran the generation to make sure nothing gets deleted or changed unexpectedly: https://github.com/mozilla/looker-hub/compare/main...test-parallel-generation
Some of the probe data was outdated when I was running the generation so some dimensions did change, and some of the descriptions did change as well. But no data got deleted.

### DIFF
--- a/generator/dryrun.py
+++ b/generator/dryrun.py
@@ -38,7 +38,8 @@ class DryRunError(Exception):
 
     def __init__(self, message, error, use_cloud_function, table_id):
         """Initialize DryRunError."""
-        super().__init__(message)
+        # super().__init__(message, error, use_cloud_function, table_id)
+        super(DryRunError, self).__init__(message, error, use_cloud_function, table_id)
         self.error = error
         self.use_cloud_function = use_cloud_function
         self.table_id = table_id
@@ -58,6 +59,8 @@ class DryRun:
 
     def __init__(
         self,
+        use_cloud_function=False,
+        id_token=None,
         sql=None,
         project="moz-fx-data-shared-prod",
         dataset=None,
@@ -67,12 +70,15 @@ class DryRun:
         """Initialize dry run instance."""
         self.sql = sql
         self.use_cloud_function = use_cloud_function
-        self.client = client
         self.project = project
         self.dataset = dataset
         self.table = table
         self.dry_run_url = dry_run_url
         self.id_token = id_token
+
+    @cached_property
+    def client(self):
+        return bigquery.Client()
 
     @cached_property
     def dry_run_result(self):
--- a/generator/dryrun.py
+++ b/generator/dryrun.py
@@ -16,6 +16,16 @@ DRY_RUN_URL = (
 )
 
 
+def credentials():
+    """Get GCP credentials."""
+    auth_req = GoogleAuthRequest()
+    creds, _ = google.auth.default(
+        scopes=["https://www.googleapis.com/auth/cloud-platform"]
+    )
+    creds.refresh(auth_req)
+    return creds
+
+
 def id_token():
     """Get token to authenticate against Cloud Function."""
     auth_req = GoogleAuthRequest()
@@ -23,6 +33,7 @@ def id_token():
         scopes=["https://www.googleapis.com/auth/cloud-platform"]
     )
     creds.refresh(auth_req)
+
     if hasattr(creds, "id_token"):
         # Get token from default credentials for the current environment created via Cloud SDK run
         id_token = creds.id_token
@@ -38,10 +49,21 @@ class DryRunError(Exception):
 
     def __init__(self, message, error, use_cloud_function, table_id):
         """Initialize DryRunError."""
-        super(DryRunError, self).__init__(message, error, use_cloud_function, table_id)
+        super().__init__(message)
         self.error = error
         self.use_cloud_function = use_cloud_function
         self.table_id = table_id
+
+    def __reduce__(self):
+        """
+        Override to ensure that all parameters are being passed when pickling.
+
+        Pickling happens when passing exception between processes (e.g. via multiprocessing)
+        """
+        return (
+            self.__class__,
+            self.args + (self.error, self.use_cloud_function, self.table_id),
+        )
 
 
 class Errors(Enum):
@@ -53,6 +75,42 @@ class Errors(Enum):
     PERMISSION_DENIED = 4
 
 
+class DryRunContext:
+    """DryRun builder class."""
+
+    def __init__(
+        self,
+        use_cloud_function=False,
+        id_token=None,
+        credentials=None,
+        dry_run_url=DRY_RUN_URL,
+    ):
+        """Initialize dry run instance."""
+        self.use_cloud_function = use_cloud_function
+        self.dry_run_url = dry_run_url
+        self.id_token = id_token
+        self.credentials = credentials
+
+    def init(
+        self,
+        sql=None,
+        project="moz-fx-data-shared-prod",
+        dataset=None,
+        table=None,
+    ):
+        """Initialize a DryRun instance."""
+        return DryRun(
+            use_cloud_function=self.use_cloud_function,
+            id_token=self.id_token,
+            credentials=self.credentials,
+            sql=sql,
+            project=project,
+            dataset=dataset,
+            table=table,
+            dry_run_url=self.dry_run_url,
+        )
+
+
 class DryRun:
     """Dry run SQL."""
 
@@ -60,6 +118,7 @@ class DryRun:
         self,
         use_cloud_function=False,
         id_token=None,
+        credentials=None,
         sql=None,
         project="moz-fx-data-shared-prod",
         dataset=None,
@@ -74,11 +133,12 @@ class DryRun:
         self.table = table
         self.dry_run_url = dry_run_url
         self.id_token = id_token
+        self.credentials = credentials
 
     @cached_property
     def client(self):
         """Get BigQuery client instance."""
-        return bigquery.Client()
+        return bigquery.Client(credentials=self.credentials)
 
     @cached_property
     def dry_run_result(self):
--- a/generator/dryrun.py
+++ b/generator/dryrun.py
@@ -58,9 +58,6 @@ class DryRun:
 
     def __init__(
         self,
-        client=None,
-        use_cloud_function=False,
-        id_token=None,
         sql=None,
         project="moz-fx-data-shared-prod",
         dataset=None,
--- a/generator/dryrun.py
+++ b/generator/dryrun.py
@@ -38,7 +38,6 @@ class DryRunError(Exception):
 
     def __init__(self, message, error, use_cloud_function, table_id):
         """Initialize DryRunError."""
-        # super().__init__(message, error, use_cloud_function, table_id)
         super(DryRunError, self).__init__(message, error, use_cloud_function, table_id)
         self.error = error
         self.use_cloud_function = use_cloud_function
@@ -78,6 +77,7 @@ class DryRun:
 
     @cached_property
     def client(self):
+        """Get BigQuery client instance."""
         return bigquery.Client()
 
     @cached_property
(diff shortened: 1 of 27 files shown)


generator/dryrun.py
```diff
@@ -38,7 +38,8 @@ class DryRunError(Exception):
 
     def __init__(self, message, error, use_cloud_function, table_id):
         """Initialize DryRunError."""
-        super().__init__(message)
+        # super().__init__(message, error, use_cloud_function, table_id)
+        super(DryRunError, self).__init__(message, error, use_cloud_function, table_id)
```
scholtzan, line 42: All arguments need to be passed, otherwise due to the use of multiprocessing a pickling error will be thrown

generator/dryrun.py
```diff
@@ -70,13 +69,17 @@ def __init__(
...
         self.dataset = dataset
         self.table = table
         self.dry_run_url = dry_run_url
         self.id_token = id_token
 
+    @cached_property
+    def client(self):
+        """Get BigQuery client instance."""
+        return bigquery.Client()
+
```
scholtzan, line 82: moving BigQuery client instance creation into the class as it's not possible to pass a client instance across different processes

generator/views/lookml_utils.py
```diff
@@ -235,7 +239,7 @@ def _is_nested_dimension(dimension: dict):
 
 def render_template(filename, template_folder, **kwargs) -> str:
     """Render a given template using Jinja."""
-    env = Environment(loader=PackageLoader("generator", f"{template_folder}/templates"))
+    env = Environment(loader=FileSystemLoader(PATH / f"{template_folder}/templates"))
```
scholtzan, line 242: Switching to file system loader here as I was getting `package was not installed in a way that PackageLoader understands` when running this using the multiprocessing pools

generator/lookml.py
```diff
@@ -129,42 +152,98 @@ def _lookml(namespaces, glean_apps, target_dir, dryrun, namespace_filter=[]):
...
+                        view,
+                        target,
+                        namespace,
+                        dryrun,
+                    )
+                )
 
-            logging.info("  Generating datagroups")
-            generate_datagroups(views, target, namespace, dryrun=dryrun)
+    with Pool(parallelism) as pool:
```
scholtzan, line 190: I tested this with `ThreadPool` as well, but using `Pool` it was about 2x faster when I tested it locally

generator/lookml.py
```diff
@@ -129,42 +152,98 @@ def _lookml(namespaces, glean_apps, target_dir, dryrun, namespace_filter=[]):
...
+                    explore_name,
+                    explore,
+                    view_dir,
+                    v1_name,
+                )
+                for explore_name, explore in explores.items()
+            ]
+
+    logging.info("  Generating explores")
+    with Pool(parallelism) as pool:
```
sean-rose, line 218: I'd suggest trying to use one pool for everything for better efficiency (no need to repeatedly tear down old processes and spin up new sets of processes for explores and dashboards), and building up the lists of things to generate in one pass would also avoid the repeated namespace loop logic.

For max efficiency you could throw everything at the pool at once, or to keep things a bit more orderly each set of artifacts could be processed in turn, like
```python
    with Pool(parallelism) as pool:
        logging.info("  Generating views")
        pool.map(run_generation, generate_views)
        logging.info("  Generating datagroups")
        pool.map(run_generation, generate_datagroups)
        ...
```

generator/dryrun.py
sean-rose: > Using the cloud function, generation completes in about 5 minutes. Using service credentials, generation takes about 10 minutes as the Bigquery client needs to be instantiated for every process as it cannot be shared across processes.

It might be worth trying to initialize one set of BigQuery credentials (which do appear to support pickling) and share those across the processes (passing them in to the `bigquery.Client` constructor's `credentials` argument) to see if that can speed things up closer to the cloud function performance, as [Google's docs](https://cloud.google.com/apis/docs/client-libraries-best-practices#reuse_client_objects_and_sessions) indicate doing the authentication, authorization, and access token generation can take multiple seconds.

generator/lookml.py
```diff
@@ -204,16 +283,39 @@ def _lookml(namespaces, glean_apps, target_dir, dryrun, namespace_filter=[]):
...
-
     glean_apps = _get_glean_apps(app_listings_uri)
 
-    dryrun = functools.partial(
-        DryRun, bigquery.Client(), use_cloud_function, id_token()
+    dry_run_id_token = None
+    if use_cloud_function:
+        dry_run_id_token = id_token()
+
+    dryrun = functools.partial(DryRun, use_cloud_function, dry_run_id_token)
```
sean-rose, line 311: Passing this partial `DryRun` constructor function across module boundaries still seems less than ideal (e.g. there's no great way to type hint it).  Maybe this could be better structured using some new class that's initialized with just the parts we want shared and has a method to do/create the dryrun (e.g. `DryRunContext` or `DryRunFactory`)?

generator/metrics_utils.py
```diff
@@ -13,7 +13,7 @@ class _MetricsConfigLoader:
     """Loads metric config files from an external repository."""
 
     config_collection: Optional[ConfigCollection] = None
-    repos: List[str] = [METRIC_HUB_REPO]
+    repos: List[str] = [METRIC_HUB_REPO, LOOKER_METRIC_HUB_REPO]
```
sean-rose, line 16: This change seems unrelated to the parallelization implementation.  Is this a bug fix, or some other intentional change in LookML generation behavior?

tests/test_lookml.py
```diff
@@ -705,6 +706,24 @@ def msg_glean_probes():
...
+    Needed when mocking objects that are used within a multiprocessing pool.
+
+    The default start method when using multiprocessing is spawn on MacOS and Windows.
+    If using fork, your processes are forked in the current state, including the mocking,
+    while with using spawn, a new Python interpreter is launched, where the mock will not work.
+    The start method needs to be changed to fork for the mocks to work across processes.
+    """
+    start_method = multiprocessing.get_start_method()
+    try:
+        multiprocessing.set_start_method("fork", force=True)
```
sean-rose, line 721: The [Python docs](https://docs.python.org/3.10/library/multiprocessing.html#contexts-and-start-methods) say that the `fork` method isn't available for Windows and should be considered unsafe on macOS.  Could we avoid the need to do this somehow? (e.g. mocking `Pool`, or special-casing `parallelism=1` to not use `multiprocessing`)

tests/test_lookml.py
```diff
@@ -705,6 +706,24 @@ def msg_glean_probes():
...
+        yield
+    finally:
+        multiprocessing.set_start_method(start_method, force=True)
```
sean-rose, line 724: The [Python docs](https://docs.python.org/3.10/library/multiprocessing.html#contexts-and-start-methods) say "`set_start_method()` should not be used more than once in the program", but it doesn't seem like their recommended alternative of using [`get_context()`](https://docs.python.org/3.10/library/multiprocessing.html#multiprocessing.get_context) would be as easy to use because it isn't a context manager that automatically takes precedence, we'd have to change the code to have a configurable `multiprocessing` parameter and use/override that.

generator/dryrun.py
scholtzan: Added a comment and changed it to override `__reduce__`

generator/lookml.py
```diff
@@ -29,82 +31,84 @@
...
+    except DryRunError as e:
+        if e.error == Errors.PERMISSION_DENIED and e.use_cloud_function:
+            print(
+                f"Permission error dry running {view.name}. Copy existing {path} file from looker-hub."
+            )
+            try:
+                get_file_from_looker_hub(path)
+            except Exception as ex:
+                print(f"Skip generating view for {path}: {ex}")
+            return path
```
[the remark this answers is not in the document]
scholtzan: mypy was complaining that no value got returned in that path. But it makes sense to return `None` in this case (even though it's not used atm)

generator/lookml.py
scholtzan: Good point. I changed the logic. It still is faster to combine running the view and data group generation as they don't depend on each other (unlike explore and dashboard generation which read view lookml).

generator/metrics_utils.py
scholtzan: Not fixing a bug, but those should be the default repos to load metrics from

tests/test_lookml.py
scholtzan: I changed the logic to use a for loop when `parallelism == 1` (which is even slightly faster)

generator/lookml.py
scholtzan: mh, it is possbile. But then calls would look something like `dryrun.create(sql=..., table=...).get_schema()`

scholtzan: I added a separate commit (the last one) with a `DryRunContext`. I think it's probably a bit more readable

scholtzan:
Just ran the generation locally and the run with GCP credentials is now very close to the cloud run. (6:30min) Though a few artifacts got skipped due to permissions.

generator/dryrun.py
```diff
@@ -53,14 +75,50 @@ class Errors(Enum):
...
+        self.credentials = credentials
+
+    def init(
+        self,
+        sql=None,
+        project="moz-fx-data-shared-prod",
+        dataset=None,
+        table=None,
+    ):
+        """Initialize a DryRun instance."""
```
sean-rose, line 101: I think naming this `init()` could be a bit confusing, as to me "init" implies modifying the object itself (the context in this case).  I'd suggest a name that sounds like it returns something (e.g. `create()`).

generator/lookml.py
```diff
@@ -118,7 +121,29 @@ def _glean_apps_to_v1_map(glean_apps):
...
 
 
-def _lookml(namespaces, glean_apps, target_dir, dryrun, namespace_filter=[]):
+def _run_generation(func):
+    """
+    Run the partially applied generate function.
+
+    For parallel execution.
+    """
+    return func()
```
sean-rose, line 130: I'd think there should be a more straightforward way to execute a bunch of ready-to-run function calls in parallel without this kind of shim, but I'm not seeing such a way. 🤷 

I guess the alternative in this case would be to build up lists of arguments to the generation functions and then parallel-map those generation functions to the arguments, but then you'd lose the flexibility to do things like combining the view and datagroup generation into one set.

generator/lookml.py
```diff
@@ -204,16 +289,46 @@ def _lookml(namespaces, glean_apps, target_dir, dryrun, namespace_filter=[]):
     help="Use the Cloud Function to run dry runs during LookML generation.",
     type=bool,
 )
+@click.option(
+    "--parallelism",
+    "-p",
+    default=8,
+    type=int,
+    help="Number of threads to use for lookml generation",
```
sean-rose, line 297:
```suggestion
    help="Number of processes to use for LookML generation",
```

generator/lookml.py
```diff
@@ -29,82 +29,85 @@
...
+            raise
 
 
-def _generate_explores(
+def _generate_explore(
     out_dir: Path,
     namespace: str,
-    explores: dict,
+    explore_name: str,
+    explore: Any,
```
sean-rose, line 70: Similar to my suggestion for `_generate_dashboard()`'s `dashboard_info` argument, if this argument were named `explore_info` then in the function body `explore_by_type` could be named `explore` like it was before without conflict.

setup.py
```diff
@@ -28,6 +28,7 @@
         ]
     },
     include_package_data=True,
+    package_data={"generator": ["dashboards/templates/*.lkml"]},
```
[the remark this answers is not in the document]
sean-rose: You reacted 👍 and marked this as resolved, but I don't see the change applied.

sean-rose (approved):
r+wc

PR #1026, opened 2024-08-02
Change www_site_downloads to shared prod

BenWu:
I think this was missed in https://github.com/mozilla/lookml-generator/pull/1025/.  lookml is currently failing with `Not found: Dataset moz-fx-data-marketing-prod:mozilla_org`


kwindau:
Thank you @BenWu , good catch

PR #1036, opened 2024-08-14
add health_issues to custom namespaces

clouserw:


clouserw:
@scholtzan I'd like to access this table in Looker.  What do you think?

scholtzan (approved):

PR #1040, opened 2024-08-21
Install spectacles used for validating lookml

scholtzan:
The plan is to use `spectacles` to run some Looker content validation via Airflow (after lookml-generator runs). The content validation can be set on a per folder level, so it finishes relatively quickly and validates dashboards and related explores. Hopefully, this way we can detect breakages faster. 

This PR installs `spectacles` as part of the `lookml-generator` docker container which will be used in Airflow. (PR to come)


kwindau (approved):
LGTM - thanks for adding this!

PR #1047, opened 2024-09-03
Update custom-namespaces.yaml

rebecca-burwei:
I would like a table from my sandbox to appear in Looker, so that we can start mocking up dashboards before the actual data is ready.


scholtzan (approved):
There is a chance that the Looker service account might not have access to `moz-fx-data-bq-data-science`, but we can try to deploy it. If it doesn't work I can figure something out.

PR #1048, opened 2024-09-03
Add task_run_costs table to fxci custom namespace

ahal:


custom-namespaces.yaml
```diff
@@ -1696,6 +1696,10 @@ fxci:
       type: table_explore
       views:
         base_view: task_runs
+    task_run_costs:
```
fbertsch, line 1699: Do you want to add measures here?

fbertsch (approved):
LGTM

custom-namespaces.yaml
ahal: I'm not sure, I don't think so? But I don't really know what I'm doing :)

Looking at the other measures defined in this file, I only see count and sum. In this case, cost is already the total for the task run. So it doesn't make sense to sum or count it *within this table explore*.

That said once this lands, I'm planning to write some explores in the spoke project that allow joining this table to the other tables in this namespace. In those cases we'll want to be able to sum cost across tasks, filtering on various attributes of tasks. Will accomplishing that require a measure defined here?

fbertsch: If you're doing further refinement of these generated explores, no need to define things here. Carry on, I'll merge this as-is

PR #1054, opened 2024-09-09
fxci: add some measures and remove unnecessary table explores

ahal:
As I learn more about Looker, I realize I needed to create some explores
in the spoke project that join across these various views. I'll be
creating an explore that joins across the three "task" tables, an
explore that joins across the two "worker" tables, and possibly an
explore that joins everything?

With these new explores, I don't think there will be any use case for
keeping table specific explores lying around.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1714,7 +1714,7 @@ fxci:
     task_run_costs:
       type: table_view
       tables:
-        - table: moz-fx-data-shared-prod.fxci_derived.task_run_costs_v1
+        - table: moz-fx-data-shared-prod.fxci.task_run_costs
       measures:
         total_cost:
           type: sum
@@ -1723,7 +1723,7 @@ fxci:
     worker_costs:
       type: table_view
       tables:
-        - table: moz-fx-data-shared-prod.fxci_derived.worker_costs_v1
+        - table: moz-fx-data-shared-prod.fxci.worker_costs
       measures:
         total_cost:
           type: sum
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1702,27 +1702,6 @@ fxci:
   owners:
     - ahalberstadt@mozilla.com
   glean_app: false
-  explores:
-    tasks:
-      type: table_explore
-      views:
-        base_view: tasks
-    task_runs:
-      type: table_explore
-      views:
-        base_view: task_runs
-    task_run_costs:
-      type: table_explore
-      views:
-        base_view: task_run_costs
-    worker_costs:
-      type: table_explore
-      views:
-        base_view: worker_costs
-    worker_metrics:
-      type: table_explore
-      views:
-        base_view: worker_metrics
   views:
     tasks:
       type: table_view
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -1774,11 +1753,26 @@ fxci:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.fxci_derived.task_run_costs_v1
+      measures:
+        total_cost:
+          type: sum
+          value_format_name: usd
+          sql: "${run_cost}"
```
ahal, line 1760: I'm not really sure if I should add this here or not. Within the `task_run_costs_v1` table, the `run_cost` field doesn't need to be aggregated as it already represents the total cost for a given run.

However, I'm making some explores that join this view to some of the other views. In those explores, we would want to aggregate this column across whatever dimensions we are grouping by. So I'm bit confused around whether this measure should be defined here or somewhere else (a new view in the spoke project??). Advice welcome!

custom-namespaces.yaml
```diff
@@ -1774,11 +1753,26 @@ fxci:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.fxci_derived.task_run_costs_v1
```
lucia-vargas-a, line 1755: Our standard in Looker is to add the user facing view -rather than the table.
```suggestion
        - table: moz-fx-data-shared-prod.fxci.task_run_costs
```

custom-namespaces.yaml
```diff
@@ -1774,11 +1753,26 @@ fxci:
...
+      measures:
+        total_cost:
+          type: sum
+          value_format_name: usd
+          sql: "${run_cost}"
     worker_costs:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.fxci_derived.worker_costs_v1
```
lucia-vargas-a, line 1764:
```suggestion
        - table: moz-fx-data-shared-prod.fxci.worker_costs
```

custom-namespaces.yaml
```diff
@@ -1774,11 +1753,26 @@ fxci:
...
+      measures:
+        total_cost:
+          type: sum
+          value_format_name: usd
+          sql: "${total_cost}"
     worker_metrics:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.fxci.worker_metrics_v1
```
lucia-vargas-a, line 1773:
```suggestion
        - table: moz-fx-data-shared-prod.fxci.worker_metrics
```

custom-namespaces.yaml
lucia-vargas-a: If the value is already aggregated, a SUM would be incorrect right? In that case my recommendation would be adding the columns as-is to looker-hub (this PR) and doing any joins/additional aggregations with different groupings in spoke-default.

lucia-vargas-a (changes requested):
Changing tables for views.

lucia-vargas-a (changes requested):
+ Ideally add a description to the PR, which can be a Jira ticket.

ahal:
Thanks for the review, fixed and added a description (I had it in a commit message but since I added multiple commits Github didn't show it)

lucia-vargas-a (approved):

PR #1059, opened 2024-09-10
Update dependencies related to spectacles

scholtzan:
The new `mozilla-nimbus-schema` version now supports recent pydantic versions, so `spectacles` can be upgraded to the most recent version


edugfilho (approved):
LGTM

PR #1074, opened 2024-10-01
DENG-2407 Allow hiding generated explores

akkomar:
This adds an option to make generated explores hidden.

This will be useful for https://mozilla-hub.atlassian.net/browse/DENG-2407 where we are deprecating custom ping in favor of `events` ping. Hiding the custom ping explore will make working with Looker less confusing to users. It will also ensure that no new dashboards are built with explore in question and existing ones will not break.

In order to hide autogenerated explore we'll need to put it in `custom-namespaces` with `hidden: true`. Specifically, for DENG-2407 this would look like:
```
accounts_backend:
  explores:
    accounts_events:
      hidden: true
      type: glean_ping_explore
      views:
        base_view: accounts_events
```

### DIFF
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -25,7 +25,9 @@ class Explore:
         """Explore instance represented as a dict."""
         return {self.name: {"type": self.type, "views": self.views}}
 
-    def to_lookml(self, v1_name: Optional[str]) -> List[Dict[str, Any]]:
+    def to_lookml(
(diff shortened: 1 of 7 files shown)


akkomar:
@scholtzan can you take a quick look at this draft PR and let me know if this feature looks good to you before I add some tests?

scholtzan:
Yeah, good idea! Looks fine to me so far

scholtzan (approved):

PR #1079, opened 2024-10-07
Add shared_private namespace for country customizations

scholtzan:
This is to fix https://mozilla.slack.com/archives/C01E8GDG80N/p1728324080900889

Currently, there are conflicting namespaces as `shared` already exists in spoke-default. The naming convention for private namespaces is to add a `_private` suffix. This will also resolve the conflict here.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1373,6 +1373,7 @@ shared_private:
   pretty_name: Shared Private
   owners:
     - ascholtz@mozilla.com
+  spoke: looker-spoke-private
   views:
     countries:
       type: table_view
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1354,7 +1354,6 @@ shared:
   glean_app: false
   pretty_name: Shared
   owners:
-    - anicholson@mozilla.com
     - ascholtz@mozilla.com
   views:
     countries:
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -1369,6 +1368,16 @@ shared:
       type: table_view
       tables:
         - table: mozdata.telemetry.applications
+shared_private:
+  glean_app: false
+  pretty_name: Shared Private
+  owners:
+    - ascholtz@mozilla.com
+  views:
```
sean-rose, line 1376:
```suggestion
  spoke: looker-spoke-private
  views:
```

custom-namespaces.yaml
```diff
@@ -1369,6 +1368,17 @@ shared:
       type: table_view
       tables:
         - table: mozdata.telemetry.applications
+shared_private:
```
sean-rose, line 1371: If they just need to reference the generated shared `countries` view then it doesn't seem like this `shared-private` spoke namespace is necessarily needed, as `looker-hub` views can be referenced in `looker-spoke-private`.

sean-rose: That said, I guess having the spoke namespace would allow for refining the `countries` view, though in that case any refinements would have to be duplicated in both `looker-spoke-default` and `looker-spoke-private` (and there are [already refinements to the `countries` view in `looker-spoke-default`](https://github.com/mozilla/looker-spoke-default/blob/main/shared/views/countries.view.lkml)).

I wonder if it would make sense to have `looker-spoke-default` referencable from `looker-spoke-private` (similar to how `looker-hub` is)?

scholtzan: I just tried. About 100 errors in validation. I guess it might not be happy about `.model` files being present in spoke-default (which aren't in looker-hub).

sean-rose (approved):
Approving so you can merge this if you decide adding the `shared_private` namespace is the right approach.

PR #1103, opened 2024-11-13
Add additional accounts health tables

clouserw:
The initial health table is working so expanding to add the additional health tables

PR #1105, opened 2024-11-13
fix lookml error due to auto-generated count column

clouserw:
related to #1103

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -610,16 +610,10 @@ firefox_accounts:
       type: table_view
       tables:
         - table: mozdata.analysis.wclouser_fxa_health_counts
-      measures:
-        count:
-          type: count
     health_db_counts:
       type: table_view
       tables:
         - table: mozdata.analysis.wclouser_fxa_health_db_counts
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -610,16 +610,10 @@ firefox_accounts:
       type: table_view
       tables:
         - table: mozdata.analysis.wclouser_fxa_health_counts
-      measures:
-        count:
-          type: count
```
sean-rose, line 615: You could still have a count measure in these views if you want, you'd just have to pick a different name for it (e.g. `row_count`):
```suggestion
        row_count:
          type: count
```

Or you could have a measure that sums the existing `count` column values ([example](https://github.com/mozilla/lookml-generator/blob/e7dd2f557e7a9c4c1870a082f87e7735cd5a0d5e/custom-namespaces.yaml#L1658-L1660)):
```suggestion
        count_sum:
          type: sum
          sql: ${count}
```

clouserw: Thanks.  Let me play with it a bit to see what works best

PR #1107, opened 2024-11-19
Bug 1932210 - Add labeled_quantity glean metric type

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1932210

Following the instructions in https://github.com/mozilla/glean/blob/main/docs/dev/core/new-metric-type/platform.md


chutten (approved):
...should I have added `labeled_{custom|memory|timing}_distribution` to this generator as well?

BenWu:
> ...should I have added labeled_{custom|memory|timing}_distribution to this generator as well?

I think so but I'm not sure how much these ping views are used

PR #1110, opened 2024-11-26
Table explores persist with datagroup

scholtzan:
This PR generates datagroups for each `TableExplore`. These datagroups determine how long results should be cached for queries based on these table explores. Caches are getting invalidated when new data is added to the underlying table. 

This is limited to table explores for now, but if this looks promising we can extend it to be used for other explores as well.


generator/views/datagroups.py
```diff
@@ -57,22 +57,35 @@ def __lt__(self, other) -> bool:
...
+            name=f"{view.name}_last_updated",
+            label=f"{view.name} Last Updated",
+            description=f"Updates for {view.name} when {full_table_id} is modified.",
+            sql_trigger=SQL_TRIGGER_TEMPLATE.format(
+                project_id=project_id,
+                dataset_id=dataset_id,
+                table_id=table_id,
+            ),
+        ),
+    ]
```
scholtzan, line 88: Ideally, this should be refactored at some point since we are currently creating duplicate datagroups. However, this was necessary to associate explores with the datagroup. The datagroups that were previously generated were named based on the BigQuery table, which the explores have no context on

kwindau (approved):
Looks good to me!

PR #1116, opened 2024-12-05
add new accounts db count table. remove old one

clouserw:
See related new table @ https://github.com/mozilla/bigquery-etl/pull/6511 . /cc @akkomar @sean-rose


custom-namespaces.yaml
```diff
@@ -571,6 +571,13 @@ firefox_accounts:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.accounts_db.fxa_oauth_clients
+    monitor_db_counts:
```
sean-rose, line 574: I'd recommend not using "monitor" in the name, as that could be confused with the Monitor product (e.g. the `monitor_mozilla_accounts_funnels` view defined below is related to the Monitor product).
```suggestion
    database_counts:
```

custom-namespaces.yaml
```diff
@@ -571,6 +571,13 @@ firefox_accounts:
...
+      type: table_view
+      tables:
+        - table: moz-fx-data-shared-prod.backend_derived.monitoring_db_counts_v1
```
sean-rose, line 577: The table reference is wrong, as the dataset should be `accounts_backend_derived`.  However, since @akkomar is adding a view specifically for this in mozilla/bigquery-etl#6599 we should wait until that gets deployed and use the view instead.
```suggestion
        - table: moz-fx-data-shared-prod.accounts_backend.monitoring_db_counts
```

custom-namespaces.yaml
```diff
@@ -571,6 +571,13 @@ firefox_accounts:
...
+      measures:
+        count:
+          type: count
```
sean-rose, line 580: This will just count the rows in the query, but that isn't very useful information. I'm guessing you want to sum the `total_rows` column values instead.
```suggestion
        total_rows_sum:
          type: sum
          sql: ${total_rows}
```
(IMO it'd be nicer if the column in the underlying table/view didn't have "total" in the name, and instead was named something like `row_count`, then this measure could be something like `total_row_count`)

custom-namespaces.yaml
```diff
@@ -694,6 +701,10 @@ firefox_accounts:
       type: table_explore
       views:
         base_view: accounts_oauth_clients
+    monitor_db_counts:
+      type: table_explore
+      views:
+        base_view: monitor_db_counts
```
sean-rose, line 707:
```suggestion
    database_counts:
      type: table_explore
      views:
        base_view: database_counts
```

custom-namespaces.yaml
```diff
@@ -610,10 +617,10 @@ firefox_accounts:
...
       tables:
         - table: mozdata.analysis.wclouser_fxa_health_counts
-    health_db_counts:
-      type: table_view
-      tables:
-        - table: mozdata.analysis.wclouser_fxa_health_db_counts
+      measures:  
+        count_sum:
+          type: sum
+          sql: ${count}
```
sean-rose, line 623: This is kind of an unrelated change, but we'll allow it. 😉

custom-namespaces.yaml
clouserw: I didn't love it, but was trying to mirror the source table name.  this wfm 👍

custom-namespaces.yaml
clouserw: 😰

custom-namespaces.yaml
sean-rose: If you'd prefer to exactly mirror the the source table name and go with `monitoring_db_counts` that'd also be reasonable to me. It's `monitor` (without the `ing`) that I think would be confusing.

And in my suggested changes I proposed `database` rather than `db` because of the way Looker auto-converts view and explore names in LookML to title case the UI, and "Database Counts" seemed preferable to "Db Counts" to me.

clouserw: I have no strong feelings.  It's under the accounts header, so, /shrug  Let's go with what we've got so we can land it

sean-rose (approved):
The `moz-fx-data-shared-prod.accounts_backend.monitoring_db_counts` view added in https://github.com/mozilla/bigquery-etl/pull/6599 has been deployed, so this is good to go.

clouserw:
Thanks.  I don't have write-access to this repository fwiw

PR #1129, opened 2024-12-20
Append `_base` to fxci view names

ahal:
I'm planning to extend these views in looker-spoke-default with some Liquid variables and custom dimensions. Appending '_base' here makes the naming scheme easier to reason about.


ahal:
I think this requires a coordinated change in looker-spoke-default to update the explore? Is there a recommended way process to land these?

ahal:
> I think this requires a coordinated change in looker-spoke-default to update the explore? Is there a recommended process to land these?

Note, I don't care if my explore is temporarily broken.. I just don't want to cause any validation errors for other folks.

scholtzan:
Yeah, this requires all of the imports from looker-hub that reference these views to be removed from spoke-default. I found this one, for example: https://github.com/mozilla/looker-spoke-default/blob/5a2d1470f303623d12e2b8919c29f05eb60d70d9/fxci/explores/tasks.explore.lkml#L2

ahal:
Ok so delete the explore, land this, re-add the explore with updated names.

I guess I could alternatively append `_extended` to the extended views.. but I feel like that will be a minor annoyance to anyone working on these in perpetuity. I think this is still worth doing.

ahal:
Here's the PR to disable the explore:
https://github.com/mozilla/looker-spoke-default/pull/1003

scholtzan (approved):

PR #1147, opened 2025-01-17
Add fxa health lighthouse scores

clouserw:
Hi. It's me.  These are front-end performance lighthouse scores, generated daily.  I'd like to put up a dashboard for them in looker.  Thanks

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -607,6 +607,13 @@ firefox_accounts:
       measures:
         count:
           type: count
+    health_lighthouse:
+      type: table_view
+      tables:
+        - table: mozdata.analysis.wclouser_fxa_health_lighthouse
+      measures:
+        count:
+          type: count
     health_pagerduty:
       type: table_view
       tables:
@@ -732,6 +739,10 @@ firefox_accounts:
       type: table_explore
       views:
         base_view: health_issues
+    health_lighthouse:
+      type: table_explore
+      views:
+        base_view: health_lighthouse
     health_pagerduty:
       type: table_explore
       views:


clouserw:
@sean-rose r?

custom-namespaces.yaml
```diff
@@ -732,6 +739,10 @@ firefox_accounts:
       type: table_explore
       views:
         base_view: health_issues
+    health_issues:
```
sean-rose, line 742: There's already a `health_issues` explore defined above.
```suggestion
    health_lighthouse:
```

custom-namespaces.yaml
```diff
@@ -607,6 +607,13 @@ firefox_accounts:
       measures:
         count:
           type: count
+    health_lighthouse:
+      type: table_view
+      tables:
+        - table: mozdata.analysis.wclouser_fxa_health_lighthouse
+      measures:
+        count:
+          type: count
```
sean-rose, line 616: Do you want to define any other measures/aggregations (e.g. for the `score` or `numVal` columns)?

custom-namespaces.yaml
```diff
@@ -607,6 +607,13 @@ firefox_accounts:
       measures:
         count:
           type: count
+    health_lighthouse:
+      type: table_view
+      tables:
+        - table: mozdata.analysis.wclouser_fxa_health_lighthouse
```
sean-rose, line 613: Note that new tables created in `mozdata.analysis` now default to expiring in 6 months, so this table is currently set to expire on July 16.  Since you already have quite a few other tables like this in `mozdata.analysis` with no table expiration which are wired up to Looker, IMO it'd be reasonable for you to [edit this table to also not expire](https://cloud.google.com/bigquery/docs/managing-tables#updating_a_tables_expiration_time).

However, I think the general idea is that going forward `mozdata.analysis` should only be used for ad hoc/short term artifacts and not long-term artifacts or things relied on by other infra like LookML.  So the preferred solution for managing such BigQuery external tables pointing to Google Sheets when they're being relied on in Looker would probably be to [define them in BigQuery ETL](https://docs.telemetry.mozilla.org/cookbooks/operational/connecting_external_data_bigquery.html#connecting-sheets) (e.g. in the `firefox_accounts_derived` dataset).

custom-namespaces.yaml
clouserw: fixed in new push. thanks

custom-namespaces.yaml
clouserw: The answer is yes, but I'm not pro enough to know what they look like in yaml yet.  I have to make them by hand in looker first and see what I use. :-/

custom-namespaces.yaml
clouserw: Thanks for checking the expiry.  I had actually fixed it the first time but had to recreate the table and forgot the next time.

I'll read the ETL docs linked.

PR #1159, opened 2025-02-06
Add Zoom views (DENG-6964)

sean-rose:
## [DENG-6964](https://mozilla-hub.atlassian.net/browse/DENG-6964): Integrate Zoom into BigQuery/Looker using FiveTran (API)

After this is deployed I plan to manually add explores joining these views together in [looker-spoke-default](https://github.com/mozilla/looker-spoke-default).


sean-rose:
The current [generate-lookml CI error](https://app.circleci.com/pipelines/github/mozilla/lookml-generator/3985/workflows/97673d7f-eb56-4872-8d9e-2a65080d75b5/jobs/19350) is likely because the views authorizations haven't been applied by data infra automation yet.  I'll retry the CI tomorrow when the view authorizations will hopefully be in place.

custom-namespaces.yaml
```diff
@@ -1960,3 +1960,26 @@ jira_service_desk:
...
         base_view: user
+zoom:
+  pretty_name: Zoom
+  connection: bigquery-oauth
+  owners:
+    - rantwan@mozilla.com
+  glean_app: false
+  views:
+    meetings:
+      type: table_view
```
scholtzan, line 1971: Does it make sense to also generate `table_explore`s for these?

sean-rose: My plan is to have one main "meetings" explore joining several of these views together, and a "users" explore just on the `users` view.  So while I could add a `table_explore` for the latter here, I figured since I'll be needing to do a looker-spoke-default PR for the other explore I'd just do both that way.

sean-rose:
> The current [generate-lookml CI error](https://app.circleci.com/pipelines/github/mozilla/lookml-generator/3985/workflows/97673d7f-eb56-4872-8d9e-2a65080d75b5/jobs/19350) is likely because the views authorizations haven't been applied by data infra automation yet. I'll retry the CI tomorrow when the view authorizations will hopefully be in place.

The `generate-lookml` CI is still [failing](https://app.circleci.com/pipelines/github/mozilla/lookml-generator/3985/workflows/d8d89b6b-ffc5-433e-8362-157a1289f643/jobs/19366) with the view authorizations in place.  It appears what's happening is since these views select from a single table LookML generator is attempting to get information about that table to generate a datagroup for the Looker view, but it doesn't have permission to access the table.

@scholtzan do you have any advice on the best way to proceed here?

Two ideas that come to mind are:
* Have LookML generator not try to generate datagroups for authorized views.  However, detecting authorized views would require changing the dryrun logic both here and in the cloud function to return tables' labels.
* Have LookML generator ignore permission errors.

scholtzan:
I thought we had some logic in place already to ignore permission errors...

scholtzan:
It should hit this: https://github.com/mozilla/lookml-generator/blob/aef6b531d48d121c1d301b7e6334fa2ec16ad057/generator/views/datagroups.py#L204

sean-rose:
> It should hit this:
> 
> https://github.com/mozilla/lookml-generator/blob/aef6b531d48d121c1d301b7e6334fa2ec16ad057/generator/views/datagroups.py#L204

Maybe the issue is that the original `DryRunError` is getting caught and a `ValueError` is getting raised with the original exception as the cause: https://github.com/mozilla/lookml-generator/blob/aef6b531d48d121c1d301b7e6334fa2ec16ad057/generator/views/datagroups.py#L146-L149

scholtzan:
This should fix it: https://github.com/mozilla/lookml-generator/pull/1161

PR #1177, opened 2025-03-13
Add device manufacturer partnerships namespace

m-d-bowerman:
Adds a namespace for the `device_manufacturer_partnerships` dataset.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -589,7 +589,7 @@ firefox_accounts:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.firefox_accounts_derived.health_counts_v1
-      measures:  
+      measures:
         count_sum:
           type: sum
           sql: ${count}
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -1995,3 +1995,15 @@ zoom:
...
       tables:
         - table: mozdata.zoom.users
+device_manufacturer_partnerships:
+  glean_app: false
+  pretty_name: Device Manufacturer Partnerships
+  owners:
+    - mbowerman@mozilla.com
+    - kik@mozilla.com
+  spoke: looker-spoke-default
+  views:
```
kik-kik, line 2005: You can also define an explore and measure here so you don't have to do it again in looker-spoke-default:

- measures example: https://github.com/mozilla/lookml-generator/blob/0f699e0bd604e972336293feea5c27f3a7688aa2/custom-namespaces.yaml#L1313-L1320
- explore example: https://github.com/mozilla/lookml-generator/blob/0f699e0bd604e972336293feea5c27f3a7688aa2/custom-namespaces.yaml#L1456-L1460

kik-kik (approved):

PR #1201, opened 2025-04-14
[fxa] add twilio

clouserw:
@sean-rose r? if you have a moment

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -636,6 +636,19 @@ firefox_accounts:
       measures:
         count:
           type: count
+    health_twilio:
+      type: table_view
+      tables:
+        - table: moz-fx-data-shared-prod.firefox_accounts_derived.health_twilio_v1
+      measures:
+        row_count:
+          type: count
+        count_sum:
+          type: sum
+          sql: ${count}
+        price_sum:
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -636,6 +636,16 @@ firefox_accounts:
       measures:
         count:
           type: count
+    health_twilio:
+      type: table_view
+      tables:
+        - table: moz-fx-data-shared-prod.firefox_accounts_derived.health_twilio_v1
+      measures:
+        count:
+          type: count
```
sean-rose, line 645: There's a `count` column in that table and a matching `count` dimension will be auto-generated, so naming this measure `count` will conflict and cause an error.

If you do want to count the rows in the table you could pick a different name, like `row_count`:
```suggestion
        row_count:
          type: count
```

Or if you want to aggregate the existing `count` column:
```suggestion
        count_sum:
          type: sum
          sql: ${count}
```

(or both, they aren't mutually exclusive)

clouserw: I added them both. thanks

clouserw:
Yes, good idea on the explore.  I added it.

sean-rose (approved):

PR #1206, opened 2025-04-29
[DENG-8386] Generate time dimensions for metric definition views

scholtzan:
https://mozilla-hub.atlassian.net/browse/DENG-8386

Previously, time or date dimensions were not added to generated metric definition views due to naming collisions when stripping suffixes from time dimension names. This will add time dimensions and ensure that there are not duplicate time dimensions being generated.

### DIFF
--- a/generator/views/metric_definitions_view.py
+++ b/generator/views/metric_definitions_view.py
@@ -152,19 +152,27 @@ class MetricDefinitionsView(View):
                 lookml_utils._generate_dimensions_from_query(query, dryrun)
             )
 
+        # to prevent duplicate dimensions, especially when working with time dimensions
+        # where names are modified potentially causing naming collisions
+        seen_dimensions = set()
         # prepare base field data for query
-        base_view_fields = [
-            {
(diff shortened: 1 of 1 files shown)


generator/views/metric_definitions_view.py
```diff
@@ -261,10 +269,7 @@ def to_lookml(self, v1_name: Optional[str], dryrun) -> Dict[str, Any]:
...
         for data_source, dimensions in base_view_dimensions.items():
             for dimension in dimensions:
-                if (
-                    dimension["name"] not in ignore_base_fields
-                    and dimension.get("type", "") != "time"
-                ):
+                if dimension["name"] not in ignore_base_fields:
                     dimension["sql"] = (
                         "${TABLE}." + f"{data_source}_{dimension['name']}"
                     )
```
mikewilli, line 275: It seems like we could do all this logic in the same loop above, but a) probably a small performance difference for something not very time-sensitive so not a big deal, and b) not necessary to change in this PR anyway.

mikewilli (approved):

PR #1218, opened 2025-05-15
Updating view name for Android SPOC reporting

chelseybeck:


custom-namespaces.yaml
```diff
@@ -852,7 +852,7 @@ ads:
     android_pocket_spocs:
```
Marlene-M-Hirose, line 852: This line needs to be revised as well

Marlene-M-Hirose (approved):

PR #1231, opened 2025-06-02
[DENG-8623] Mechanism to exclude generated LookML content

scholtzan:
https://mozilla-hub.atlassian.net/browse/DENG-8623

Based on proposal: https://docs.google.com/document/d/1CJq_avwso-fK7F4H3483brw8yAYFA0VQbgrfOA5ZY3s/edit?tab=t.0

Our Looker instance has a lot of generated LookML, some of which is very rarely or never being looked at. lookml-generator automatically generates a set of views and explores for every Glean application. The amount of LookML content is impacting the overall performance of our Looker instance. 

This extends the existing namespaces-disallowlist.yaml to exclude views and explores from being generated.

### DIFF
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -96,7 +96,11 @@ class Explore:
     def get_view_lookml(self, view: str) -> dict:
         """Get the LookML for a view."""
         if self.views_path is not None:
-            return lkml.load((self.views_path / f"{view}.view.lkml").read_text())
+            try:
+                return lkml.load((self.views_path / f"{view}.view.lkml").read_text())
+            except Exception as e:
+                print(e)
+                print(view)
         raise Exception("Missing view path for get_view_lookml")
 
     def get_datagroup(self) -> Optional[str]:
--- a/generator/explores/explore.py
+++ b/generator/explores/explore.py
@@ -96,11 +96,8 @@ class Explore:
     def get_view_lookml(self, view: str) -> dict:
         """Get the LookML for a view."""
         if self.views_path is not None:
-            try:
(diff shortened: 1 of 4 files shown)


kwindau (approved):

generator/explores/explore.py
```diff
@@ -96,7 +96,11 @@ def from_dict(name: str, defn: dict, views_path: Path) -> Explore:
     def get_view_lookml(self, view: str) -> dict:
         """Get the LookML for a view."""
         if self.views_path is not None:
-            return lkml.load((self.views_path / f"{view}.view.lkml").read_text())
+            try:
+                return lkml.load((self.views_path / f"{view}.view.lkml").read_text())
+            except Exception as e:
+                print(e)
+                print(view)
```
sean-rose, line 103: Are such exceptions going to happen for every excluded view?

generator/namespaces.py
```diff
@@ -400,22 +401,49 @@ def namespaces(
...
+
+
+def _filter_disallowed(namespaces, disallowlist):
+    """Filter models, explores and views from the generated namespaces config, based on the disallowlist."""
+
+    def match_any(patterns, name):
+        return any(fnmatch.fnmatch(name, p) for p in patterns)
+
+    # transform namespace disallowlist to a dict
+    disallowed_namespaces = yaml.safe_load(disallowlist.read()) or {}
```
sean-rose, line 422: If the YAML gets loaded the top-level object is a list, so I'd have the fallback value be a list as well.
```suggestion
    disallowed_namespaces = yaml.safe_load(disallowlist.read()) or []
```

PR #1235, opened 2025-06-04
feat: add glean-only experimentation views

mikewilli:
Adds views needed to create glean-only dashboard for experiment enrollments monitoring.


mikewilli:
>I think one of the CI errors is because view deployment is broken right now. I'm not sure about the rest

Yea I thought the same. I saw a bunch of view deployment tasks queued up so going to wait and see if this resolves itself..

PR #1243, opened 2025-06-13
Exclude unused explores and views

scholtzan:
Exclude unused views and explores based on https://mozilla.cloud.looker.com/dashboards/2375?Submission+Date=90+day
These views and explores haven't been used in over a year and aren't on any dashboards.

Tested on https://github.com/mozilla/looker-hub/compare/ascholtz-remove-content-5?expand=1 with lookml validation succeeding

Depends on https://github.com/mozilla/looker-spoke-default/pull/1090


generator/views/glean_ping_view.py
```diff
@@ -52,7 +52,7 @@
 
 # List of labeled counter names for which a suggest explore should be generated.
 # Generating suggest explores for all labeled counters slows down Looker.
-SUGGESTS_FOR_LABELED_COUNTERS = {"metrics__labeled_counter__glean_error_invalid_label"}
```
scholtzan, line 55: These suggest explores have not been used in the last year. We don't need to generate them

namespaces-disallowlist.yaml
```diff
@@ -29,45 +29,131 @@
...
+      - home
+      - logins_sync
+      - new_metric_capture_emulation
+      - nimbus
       - pageload
+      - pocket_button
+      - startup_timeline
+      - tabs_sync
       - usage_deletion_request
+      - usage_reporting_clients_*
```
BenWu, line 159: Isn't `usage_reporting` somewhat new and used for DAU-like things?  I'm wondering if there are plans to use it but I have no idea

BenWu (approved):

namespaces-disallowlist.yaml
scholtzan: mh not sure... I guess we can add it back if it's needed

PR #1258, opened 2025-07-11
feat: delete fpa, mozilla_vpn_private, and newtab namespaces as they are empty

kik-kik:
# feat: delete fpa, mozilla_vpn_private, and newtab namespaces as they are empty


BenWu:
I just noticed the CI failure.  Not sure what that's from

PR #1260, opened 2025-07-14
DENG-9108 Don't generate explore joins for dual labeled counter metric

akkomar:


BenWu:
https://mozilla-hub.atlassian.net/browse/DENG-9108

PR #1327, opened 2025-10-24
fix: delete funnel_analysis and events_daily_table as both use teleme…

kik-kik:
…try.events_daily which has been deleted


sean-rose:
For posterity, this was a side effect of https://github.com/mozilla/bigquery-etl/pull/8324 "_DENG-9951 - Unschedule messaging_system events_daily tables_", which also removed the `telemetry.event_types` and `telemetry.events_daily` views.

PR #1353, opened 2025-11-21
DENG-8698 add tables for Data Health Scorecards - Compliance

edugfilho:
Adds shredder monitoring tables and explores under the `monitoring` namespace to be used by the `Compliance` dashboard of Data Health Scorecards.

### DIFF
--- a/custom-namespaces.yaml
+++ b/custom-namespaces.yaml
@@ -1263,6 +1263,18 @@ monitoring:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.monitoring_derived.shredder_targets_new_mismatched_v1
+    shredder_run_stats:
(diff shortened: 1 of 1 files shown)


custom-namespaces.yaml
```diff
@@ -1392,6 +1392,14 @@ monitoring:
       type: table_view
       tables:
         - table: moz-fx-data-shared-prod.monitoring.looker_usage_models
+    shredder_run_stats:
```
BenWu: Can you move these up to where the existing shredder views are defined so it's easier to see what exists?  Also can you add `moz-fx-data-shared-prod.monitoring.shredder_per_table_stats` while you're here?

BenWu (approved):

PR #1357, opened 2025-11-24
refactor: Change labels for nested fields to more clearly indicate their hierarchy with colons

sean-rose:
And also to more clearly indicate the hierarchy of metric categories and names.

You can see the impact of these changes in the `lookml.diff`.


scholtzan (approved):
Nice!

PR #1384, opened 2026-02-02
add forecast_suggest_monthly view and explore

jaredsnyder:
Ads a view and explore of the `ads.forecast_suggest_monthly` table


mozza-wins (approved):
lfgggggg

PR #1389, opened 2026-02-11
make metrics into measures

mozza-wins:
using the example that Anna shared, creating measures so that we don't have to redefine them


jaredsnyder (approved):
lets go

PR #1390, opened 2026-02-11
rename measure

mozza-wins:


jaredsnyder (approved):
lgtm

Nit but it might be clearer if the connection between the measures and dimensions were more clear, IE budget is the dimension and budget_sum is the measure

PR #1427, opened 2026-04-08
feat: add experiment SRM values as view

mikewilli:
Follow-up to https://github.com/mozilla/bigquery-etl/pull/9167, which added a new table to precompute SRM for all experiments.

We can follow this by updating the SRMs on the experiment enrollments monitoring dashboard to use this view.


scholtzan (approved):
Should we add the explore for it as part of this PR too?

PR #1433, opened 2026-04-29
Remove unused prototype feature configurations

scholtzan:
I believe this workstream has been discontinued. We should be able to remove these from Looker as the analysis tables have expired.

This will unblock Looker Airflow deploys.


chelseatroy (approved):
Yep, this is great! Thanks for doing this. I can go into looker and delete the dashboard that used these also.

PR #1441, opened 2026-05-15
[DENG-11122] Don't skip generation of deprecated apps

scholtzan:
https://mozilla-hub.atlassian.net/browse/DENG-11122 

We had a recent case where deprecating `mozilla_vpn` resulted in various artifacts not being generated anymore, which broke lookml validation and existing . See https://github.com/mozilla/lookml-generator/pull/1438 

This PR removes the logic that skips generation for deprecated apps. Instead these apps need to leverage the `disallow-namespaces.yaml` after going through the deprecation process. 

Depends on https://github.com/mozilla/bigquery-etl/pull/9374

### DIFF
--- a/namespaces-disallowlist.yaml
+++ b/namespaces-disallowlist.yaml
@@ -1,10 +1,12 @@
 ---
 - burnham
+- bergamot
 - experimenter_backend
 - firefox_echo_show
 - firefox_fire_tv
 - firefox_reality
 - firefox_reality_pc
+- firefox_translations
 - lockwise_android
 - lockwise_ios
 - mlhackweek_search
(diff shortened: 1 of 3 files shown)


namespaces-disallowlist.yaml
```diff
@@ -1,10 +1,12 @@
 ---
 - burnham
+- bergamot
 - experimenter_backend
 - firefox_echo_show
 - firefox_fire_tv
 - firefox_reality
 - firefox_reality_pc
+- firefox_translations
```
scholtzan, line 9: these apps are deprecated

BenWu (approved):

PR #1448, opened 2026-07-17
added forecast_content_daily_aggregates view and explore

jaredsnyder:
Added view and explore with some measures


florezita (approved):
lgtm

